PR #23, opened 2023-01-24
Refactor/user admin page react

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3066314&fileview=grid

Signed-off-by: romanetar <roman_ag@hotmail.com>


resources/js/components/rich_text_editor.js
```diff
@@ -0,0 +1,37 @@
...
+        if (handleChangeDebounce) handleChangeDebounce.cancel()
+        handleChangeDebounce = debounce(() => {
+            const html = stateToHTML(state.getCurrentContent());
+            onChange(html);
+        }, 300);
+        handleChangeDebounce();
+    }
+
+    return (
+        <MUIRichTextEditor
```
smarcet, line 19: we need to support MD

resources/js/components/rich_text_editor.js
```diff
@@ -0,0 +1,37 @@
...
+import {stateToHTML} from 'draft-js-export-html';
+import debounce from 'lodash.debounce';
+
+const RichTextEditor = ({rteRef, rteProps, value, onChange}) => {
+    let handleChangeDebounce;
+
+    const handleChange = (state) => {
+        if (handleChangeDebounce) handleChangeDebounce.cancel()
+        handleChangeDebounce = debounce(() => {
+            const html = stateToHTML(state.getCurrentContent());
```
smarcet, line 12: bio text is not HTML , is MD

smarcet: lets take a look to https://www.npmjs.com/package/react-simplemde-editor

smarcet (changes requested):
@romanetar please review comments

resources/views/profile.blade.php
```diff
@@ -1,394 +1,131 @@
...
+            realm: "{!! $site->getRealm() !!}",
+            trusted_data: "{!! $site->getUITrustedData() !!}",
+            deleteURL: "{!! URL::action("UserController@deleteTrustedSite",["id"=>$site->getId()]) !!}"
         });
+        @endforeach
+
+        const initialValues = {
+            address1: '{!! $user->address1 ?? '' !!}',
+            address2: '{!! $user->address2 ?? '' !!}',
+            bio: '{!! $user->bio ?? '' !!}',
```
smarcet, line 27: this is breaking bc /r/n sequences
i do think that perhaps we should do the json dump at php side

resources/js/components/rich_text_editor.js
romanetar: done

resources/js/components/rich_text_editor.js
romanetar: done

resources/views/profile.blade.php
romanetar: done

app/Http/Controllers/Api/APICRUDController.php
```diff
@@ -138,8 +138,11 @@ protected function onUpdate($id, $payload){
     protected function _update($id, array $payload)
     {
         try {
+            if (array_key_exists("groups", $payload)) {
```
smarcet, line 141: what is the purpose of this ?
this controller is an abstract one not an specialized one

smarcet (changes requested):
@romanetar please review comments

resources/views/profile.blade.php
smarcet: i do think that probably the cleaner way would be using a serializer from php controller method to serialize to json the initial value of the user object

app/Http/Controllers/Api/APICRUDController.php
romanetar: Moved to UserApiController but we need this explicit convertion on server-side bc even though we send an int array the API receives a string array representation (HTTP protocol works in that way)

resources/views/admin/edit-user.blade.php
```diff
@@ -1,248 +1,125 @@
...
+    <script>
+        const current_groups = [];
+        @foreach($user->getGroups() as $group)
+        current_groups.push({"id": {!!$group->id!!}, "name": "{!!$group->name!!}"});
+        @endforeach
 
-        var current_language = '{!!$user->language!!}';
-        var current_country  = '{!!$user->country_iso_code!!}';
-        var current_gender   = '{!!$user->gender !!}';
+        const initialValues = {
```
smarcet, line 18: could we create a serializer for this to avoid this 
and serialize from controller method?

resources/views/profile.blade.php
```diff
@@ -1,394 +1,131 @@
...
+        trustedSites.push({
+            id: "{!! $site->getId() !!}",
+            auth_policy: "{!! $site->getAuthorizationPolicy() !!}",
+            realm: "{!! $site->getRealm() !!}",
+            trusted_data: "{!! $site->getUITrustedData() !!}",
+            deleteURL: "{!! URL::action("UserController@deleteTrustedSite",["id"=>$site->getId()]) !!}"
         });
+        @endforeach
+
+        const initialValues = {
```
smarcet, line 24: could we create a serializer for this to avoid this 
and serialize from controller method?

smarcet (changes requested):
@romanetar please review comments

resources/views/profile.blade.php
romanetar: done

resources/views/admin/edit-user.blade.php
romanetar: done

routes/web.php
```diff
@@ -184,6 +184,7 @@
         Route::get('', "UserApiController@getAll");
         Route::post('', ['middleware' => ['openstackid.currentuser.serveradmin.json'], 'uses' => "UserApiController@create"]);
         Route::put('me', "UserApiController@updateMe");
+        Route::get('user_actions', ['middleware' => ['openstackid.currentuser.serveradmin.json'], 'uses' => "UserActionApiController@getActions"]);
```
smarcet, line 187: use of this middle-ware is incorrect
u just use just "auth" middle ware

app/Http/Controllers/Api/UserActionApiController.php
```diff
@@ -0,0 +1,122 @@
...
+        ];
+    }
+
+    /**
+     * @return array
+     */
+    protected function getFilterValidatorRules(): array
+    {
+        return [
+            'owner' => 'nullable|int',
```
smarcet, line 73: this filter should be owner_id and required

app/Http/Controllers/Api/UserApiController.php
```diff
@@ -231,17 +231,32 @@ public function updateMe(LaravelRequest $request)
...
+        return $this->_update($myId, $payload);
+    }
+
+    /**
+     * @param $id
+     * @return \Illuminate\Http\JsonResponse|mixed
+     */
+    public function update($id)
+    {
+        $payload = $this->getUpdatePayload();
```
smarcet, line 244: why doing this normalization here ?
normalization should be done at client ( action before sending the json payload)
and u should add a validation payload rule like
groups => 'sometime|int_array'

app/Http/Controllers/Api/UserActionApiController.php
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    protected function serializerType(): string
+    {
+        return SerializerRegistry::SerializerType_Private;
+    }
+
+    protected function applyExtraFilters(Filter $filter): Filter
+    {
+        if (!$filter->hasFilter("owner")) {
```
smarcet, line 107: this is not the correct treatment for the authz part
owner ( owner_id from now on) should be required
always
u should check for the filter value ( owner_id ) at early stage of controller method
and perform authz check
if owner_id == to current user id then pass
if owner_id <> current_user_id and current user is not admin fail

smarcet (changes requested):
@romanetar please review comments

routes/web.php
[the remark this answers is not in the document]
romanetar: done

romanetar: done

app/Http/Controllers/Api/UserActionApiController.php
romanetar: done

app/Http/Controllers/Api/UserActionApiController.php
romanetar: moved ahead to early stage of controller

app/Http/Controllers/Api/UserApiController.php
romanetar: as we discussed, it stays like this

smarcet (approved):
LGTM

PR #77, opened 2025-06-16
fix: remove otp settings for service account app types

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3815497&stafilter=NotComplete&fileview=grid

### DIFF
--- a/app/Models/OAuth2/Client.php
+++ b/app/Models/OAuth2/Client.php
@@ -1673,11 +1673,13 @@ class Client extends BaseEntity implements IClient
 
     public function enablePasswordless(): void
     {
-        if (in_array($this->getApplicationType(), self::$allowed_otp_client_types)) {
-            $this->otp_enabled = true;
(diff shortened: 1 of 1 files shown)


app/Models/OAuth2/Client.php
```diff
@@ -1667,9 +1673,11 @@ public function isPasswordlessEnabled(): bool
 
     public function enablePasswordless(): void
     {
-        $this->otp_enabled = true;
-        $this->otp_length = intval(Config::get("otp.length"));
-        $this->otp_lifetime = intval(Config::get("otp.lifetime"));
+        if (in_array($this->getApplicationType(), self::$allowed_otp_client_types)) {
```
smarcet, line 1676: if app type is not valid throw validation exception

smarcet (approved):
LGTM

PR #79, opened 2025-06-18
feat: don't allow to emit otp when user isn't active

romanetar:
ref https://tipit.avaza.com/project/view#!tab=task-pane&groupby=MyTaskDueDate&view=vertical&task=3818782&stafilter=NotComplete&fileview=grid

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -306,7 +306,7 @@ final class UserController extends OpenIdController
             $user = $this->auth_service->getUserByUsername($username);
 
             if (!$user->isActive())
-                throw new ValidationException("user is not active.");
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -303,6 +303,11 @@ public function emitOTP()
                 throw new ValidationException("empty username.");
             }
 
+            $user = $this->auth_service->getUserByUsername($username);
+
+            if (!$user->isActive())
+                throw new ValidationException("user is not active.");
```
smarcet, line 309: lets change to **User** ( Uppercase )

smarcet (changes requested):
@romanetar

PR #88, opened 2025-08-07
feat: new api v2 and get user by id v2 endpoint

romanetar:
ref https://tipit.avaza.com/project/view/111133#!tab=task-pane&task=3848886


smarcet (approved):
LGTM

PR #112, opened 2026-02-12
Feat | Add OpenAPI documentation for OAuth2RocketChatSSOApiController v1

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b8e6k1y

### DIFF
--- a/app/Http/Controllers/Api/OAuth2/OAuth2RocketChatSSOApiController.php
+++ b/app/Http/Controllers/Api/OAuth2/OAuth2RocketChatSSOApiController.php
@@ -17,6 +17,9 @@ use models\exceptions\EntityNotFoundException;
 use models\exceptions\ValidationException;
 use OAuth2\IResourceServerContext;
 use Utils\Services\ILogService;
+use App\libs\OAuth2\IUserScopes;
+use OpenApi\Attributes as OA;
+use Symfony\Component\HttpFoundation\Response as HttpResponse;
 /**
  * Class OAuth2RocketChatSSOApiController
  * @package App\Http\Controllers\Api\OAuth2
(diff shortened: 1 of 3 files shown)


martinquiroga-exo (approved):
LGTM

caseylocker (changes requested):
The RocketChatUserProfileSchema is created but never referenced from the endpoint's HTTP_OK response — the response uses a bare new OA\JsonContent() instead. An orphaned empty schema adds noise to the generated spec  without providing value to API consumers.

Either wire it up by adding ref: '#/components/schemas/RocketChatUserProfile' to the JsonContent and marking the schema with additionalProperties: true (since the response proxies Rocket Chat's dynamic payload), or delete RocketChatUserProfileSchema.php entirely and keep the descriptive comment on the empty JsonContent.

app/Http/Controllers/Api/OAuth2/OAuth2RocketChatSSOApiController.php
```diff
@@ -33,31 +37,68 @@ public function __construct
...
+            ),
+        ],
+        responses: [
+            new OA\Response(
+                response: HttpResponse::HTTP_OK,
+                description: 'OK, returns Rocket Chat user profile data on login success',
+                content: new OA\JsonContent(
+                    // The content of the response is defined by "data" portion of
+                    // the Rocket Chat login endpoint response structure
+                    red: '#/components/schemas/RocketChatUserProfile',
```
caseylocker: Comeon. "ref:"

caseylocker (changes requested):
See "red:" -> "ref:" comment.

For the empty rocket chat schema property just add a flag for additionalproperties so that it isn't empty. We don't control it but we don't want to just kick the can.

```
#[OA\Schema(
      schema: 'RocketChatUserProfile',
      type: 'object',
      additionalProperties: true,
      description: 'Rocket Chat SSO user profile. The response structure is the "data" portion of the Rocket Chat /api/v1/login endpoint response and is defined by the external Rocket Chat server.'
  )]
```

User's need to know that at least there are properties and they should refer to the docs for them.

app/Http/Controllers/Api/OAuth2/OAuth2RocketChatSSOApiController.php
matiasperrone-exo: fixed

matiasperrone-exo:
Thanks @caseylocker for the comments. Now is ready to review again.
The typo "red" instead of "ref" and the suggested description were incorporated.

caseylocker (approved):
LGTM

matiasperrone-exo:
Rebased

smarcet (approved):
LGTM

PR #115, opened 2026-02-28
fix: test environment setup for clean Docker installs

JpMaxMan:
## What

Two fixes that make the test suite work correctly in clean Docker environments (i.e. when someone clones the repo fresh and runs the tests for the first time).

### 1. Fix: seed `groups` API and scopes in `TestSeeder`

`OAuth2GroupApiTest` requests the `groups/read-all` and `groups/write` scopes during its OAuth2 flow. These scopes are created by a data migration (`Version20250731141145`), but `OpenStackIDBaseTestCase` wipes `oauth2_api_scope` before every test run and re-seeds from `TestSeeder` — which did not include these scopes.

In a clean environment where migrations are run before the tests, the migration is already recorded as applied, so it does not re-run during the test, and the scopes are never restored. Result: 400 on the token endpoint, test fails.

Fixed by adding the `groups` API and its two scopes to `TestSeeder` directly, following the same pattern used for other APIs.

### 2. Docs: add `.env.testing.example` and testing instructions

When PHPUnit runs, it sets `APP_ENV=testing` (via `phpunit.xml`), causing Laravel to load `.env.testing` instead of `.env`. This is not documented anywhere, and the missing file produces no clear error — it just silently falls back to `.env`, which is missing many vars the OAuth2 server depends on.

Added:
- `.env.testing.example` — pre-configured for the standard Docker Compose setup, ready to copy
- A "Running the Test Suite" section in `LOCAL_DEVELOPMENT_HOWTO.md` explaining the `.env.testing` requirement and the `stopOnFailure` behavior
- Updated `.gitignore` to allow `.env.testing.example` to be committed despite the `.env.*` rule

## Why

Both issues caused `OAuth2GroupApiTest::testGetAll` to fail (returning 400 instead of 200) in any clean Docker setup. The root causes were non-obvious and would affect any new contributor setting up the project for the first time.

### DIFF
--- a/.env.testing.example
+++ b/.env.testing.example
@@ -1,0 +1,84 @@
+# .env.testing.example
+# Copy this to .env.testing before running the test suite.
+# Pre-configured for the standard Docker Compose local setup.
+# Adjust DB/Redis credentials if you changed them from the defaults.
+
+APP_ENV=testing
+APP_DEBUG=true
+# Generate with: php artisan key:generate --show
+APP_KEY=
+APP_URL=http://localhost
+
+DB_CONNECTION=mysql
+DB_HOST=db
+DB_PORT=3306
+DB_DATABASE=idp_local
+DB_USERNAME=idp_user
+DB_PASSWORD=1qaz2wsx!
+DB_USE_SSL=false
+
+REDIS_HOST=redis
+REDIS_PORT=6379
+REDIS_DB=0
+REDIS_PASSWORD=1qaz2wsx!
+REDIS_DATABASES=16
+
+CACHE_DRIVER=redis
+SESSION_DRIVER=redis
+SESSION_COOKIE_DOMAIN=localhost
+SESSION_COOKIE_SECURE=false
+
+QUEUE_DRIVER=redis
+QUEUE_CONN=openstackid
+QUEUE_DATABASE=openstackid
+
+MAIL_DRIVER=log
+MAIL_FROM_EMAIL=noreply@localhost
+MAIL_FROM_NAME=OpenStackID
+
+CORS_ALLOWED_HEADERS="origin, content-type, accept, authorization, x-requested-with"
+CORS_ALLOWED_METHODS="GET, POST, OPTIONS, PUT, DELETE"
+CORS_USE_PRE_FLIGHT_CACHING=true
+CORS_MAX_AGE=3200
+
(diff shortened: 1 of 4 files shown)


smarcet:
@JpMaxMan 

 The "fix" is a duplicate of existing code                                                                                                                                                         
                                                                                                                                                                                                    
The PR claims that TestSeeder doesn't seed the groups API and its scopes, causing OAuth2GroupApiTest to fail. This is wrong. The seeders already handle it:                                       
                                                                                                                                                                                                  
1. ApiSeeder (called at TestSeeder.php:448 via $this->call(ApiSeeder::class) see https://github.com/OpenStackweb/openstackid/blob/173f281d1d042261f947a70bf9502b617d2968a9/database/seeds/TestSeeder.php#L448 ) already creates the groups API at ApiSeeder.php:67-74 see https://github.com/OpenStackweb/openstackid/blob/173f281d1d042261f947a70bf9502b617d2968a9/database/seeds/ApiSeeder.php#L67 .
2. ApiScopeSeeder (called at TestSeeder.php:449 see https://github.com/OpenStackweb/openstackid/blob/173f281d1d042261f947a70bf9502b617d2968a9/database/seeds/TestSeeder.php#L449) already has a seedGroupScopes() method at ApiScopeSeeder.php:153-173 see https://github.com/OpenStackweb/openstackid/blob/71f869e79a90bf797393fe4c55d2cf843f410dab/database/seeds/ApiScopeSeeder.php#L157 that seeds both **IGroupScopes::ReadAll and IGroupScopes::Write**.
3. ApiEndpointSeeder (called at TestSeeder.php:450 see https://github.com/OpenStackweb/openstackid/blob/173f281d1d042261f947a70bf9502b617d2968a9/database/seeds/TestSeeder.php#L450 ) already has seedGroupEndpoints() at ApiEndpointSeeder.php:210-222.

The full TestSeeder::run() flow calls these three seeders before the test-specific seedApis()/seedApiScopes() methods. **The groups data is already seeded and nothing downstream deletes it.**

also 

The PR adds a groups entry to TestSeeder::seedApis(), but that method creates APIs under "test resource server", while ApiSeeder creates the groups API under "OpenStackId server". This would
create a second groups API under a different resource server.

i am gonna decline this PR  , if you want the .env.testing docs, that could be a separate small PR.

PR #116, opened 2026-03-04
Hash cache key so reserved characters do not cause keys to be rejected.

mulldug:
Cache keys were being rejected because they contained  characters that are reserved in PSR-6 cache keys.  The fix is to hash the key.

### DIFF
--- a/app/Repositories/DoctrineOAuth2ClientRepository.php
+++ b/app/Repositories/DoctrineOAuth2ClientRepository.php
@@ -122,7 +122,7 @@ final class DoctrineOAuth2ClientRepository
         $q = $qb->getQuery();
 
         $q->useQueryCache(true);
-        $q->enableResultCache(600, 'client_by_id_'.$client_id); // TTL 10 min
+        $q->enableResultCache(600, 'client_by_id_'.md5($client_id)); // TTL 10 min
         $q->setHint(Query::HINT_READ_ONLY, true);
 
         return $q->getOneOrNullResult();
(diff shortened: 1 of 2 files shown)


smarcet (commented):
Good fix — the `md5()` hashing correctly eliminates PSR-6 reserved characters (`/` in particular) from cache keys.

However, this PR is missing a regression test. Without one, the `md5()` call could be accidentally reverted with no test failure to catch it.

Suggested test coverage (integration test using the real cache backend):

1. **Cache write/read with reserved characters** — call `getClientByIdCacheable` with a client ID containing `/` (e.g. `.-_~87D8/Vcvr6fvQbH4HyNgwTlfSyQ3x.openstack.client` from `TestSeeder`), assert it returns the correct client without throwing `InvalidArgumentException`. Call it a second time to verify the cached entry is also retrievable.

2. **Distinct cache entries for different IDs** — call `getClientByIdCacheable` with two different client IDs that both contain `/`, assert they return different clients. This confirms the hashing doesn't accidentally collapse distinct keys.

These can go in a new `tests/OAuth2ClientCacheTest.php` extending `BrowserKitTestCase` (which seeds the test data and connects to Redis).

smarcet:
@mulldug please
1. we do follow Conventional Commits messages ( see https://www.conventionalcommits.org/en/v1.0.0/)
2. please provide unit test to verify the fix and add it to the CI

smarcet (changes requested):
@mulldug please review comments

mulldug:
Tests have been added and the commit message updated.

smarcet (approved):
LGTM

PR #121, opened 2026-03-24
Feat | Migrate from Google reCAPTCHA to Cloudflare Turnstile

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b8xw6ue

## Related Task (Blocks):

Ref: https://app.clickup.com/t/86b9v1acw (Harden login_attemps in server side)
PR: [#132 - Feat | Harden login captcha gate by replacing request body counter with server-side session counter](https://github.com/OpenStackweb/openstackid/pull/132)

## Goal

Replace the legacy Google reCAPTCHA v2 implementation (greggilbert/recaptcha + react-google-recaptcha) with Cloudflare Turnstile across all 6 authentication forms (login, registration, forgot
password, reset password, set password, email verification) in OpenStackID. This is a hard replacement.

## Tasks

1. Remove greggilbert/recaptcha PHP package and its custom VCS repository from composer.json.
2. Install ryangjchandler/laravel-cloudflare-turnstile.
3. Remove RecaptchaServiceProvider and Recaptcha facade from config/app.php.
4. Delete config/recaptcha.php. Publish config/turnstile.php.
5. Update .env.example (RECAPTCHA_PUBLIC_KEY/RECAPTCHA_PRIVATE_KEY → TURNSTILE_SITE_KEY/TURNSTILE_SECRET_KEY).
6. Update validation rules in RegisterController, ResetPasswordController, PasswordSetController, EmailVerificationController, and UserController: change 'g-recaptcha-response' => 'required|recaptcha' to 'cf-turnstile-response' => 'required|turnstile'.
7. Add server-side Turnstile validation to ForgotPasswordController (previously had no server-side captcha check).
8. Remove react-google-recaptcha from package.json, install @marsidev/react-turnstile.
9. Update all 5 functional React components (signup, forgot_password, reset_password, set_password, email_verification): replace ReCAPTCHA import/component with Turnstile, change sitekey to siteKey, onChange to onSuccess, captcha.current.getValue() to captcha.current?.getResponse().
10. Update the class-based login.js component: same import/prop/callback changes as above, applied to both captcha instances (password flow and OTP flow).
11. Preserve the conditional captcha display logic (shouldShowCaptcha() based on failed login attempts).
12. Update all 6 Blade templates: change Config::get("recaptcha.public_key") to Config::get("services.turnstile.site_key").
13. Rename .recaptcha to .turnstile in 5 SCSS modules.
14. Add .turnstile class to forgot_password.module.scss (was missing).
15. Delete 5 dead legacy jQuery files under public/assets/js/auth/.
16. Run full webpack build.

## Acceptance Criteria

1.  No references to greggilbert/recaptcha, react-google-recaptcha, g-recaptcha-response, or Config::get("recaptcha.\*") remain in production code
2.  composer install and yarn install && yarn build succeed
3.  All 6 controllers validate cf-turnstile-response using the turnstile rule
4.  All 6 React components render the Turnstile widget and validate the token before form submission
5.  Login form still conditionally shows the captcha widget only after N failed login attempts
6.  .env.example contains TURNSTILE_SITE_KEY and TURNSTILE_SECRET_KEY

## Dev Notes

1. The ryangjchandler/laravel-cloudflare-turnstile package auto-discovers its service provider, do NOT manually register it in config/app.php.
2. Turnstile's React component API differs from reCAPTCHA: siteKey (camelCase, not sitekey), onSuccess (not onChange), ref.current.getResponse() (not .getValue()).
3. The Turnstile widget auto-renders a hidden cf-turnstile-response input inside the form, same pattern as reCAPTCHA's g-recaptcha-response. Standard HTML form POST picks it up automatically.
4. login.js is a class component (the other 5 are functional). It has two <ReCAPTCHA> instances (password flow line ~187, OTP flow line ~266) : both need updating.
5. ForgotPasswordController::validator() currently only validates email. This migration intentionally adds 'cf-turnstile-response' => 'required|turnstile' to harden it.
6. The forgot_password.module.scss never had a .recaptcha class (the component referenced it but it resolved to undefined). Add a .turnstile class with the standard margins (10px top/bottom, text-align center) to match other forms.
7. The 5 jQuery files under public/assets/js/auth/ are dead code : not loaded by any Blade template. Safe to delete.
8. Deployment requires replacing env vars: remove RECAPTCHA_PUBLIC_KEY/RECAPTCHA_PRIVATE_KEY, add TURNSTILE_SITE_KEY/TURNSTILE_SECRET_KEY (from Cloudflare Dashboard > Turnstile).

## Changes to the original request:

1. On validation instead of using 
```php
'cf-turnstile-response' => 'required|turnstile',
```
we need to use:
```php
'cf-turnstile-response' => ['required', new Turnstile()],
```
As [documented](https://github.com/ryangjchandler/laravel-cloudflare-turnstile#usage): 
image

2. The Dev Note 7, mentions: _The 5 jQuery files under `public/assets/js/auth/` are dead code_, but there are 7 files (rather than only 5) present in the `public/assets/js/auth/` folder, of which 5 files which are actual dead code, 2 more files that are being used, and, more important, 1 file (`public/assets/js/auth/email-verification-complete.js`) which is being referenced by the method `Auth\EmailVerificationController@verify` (endpoint: `auth/verification/{token}`) inside the template `resources/views/auth/email_verification_success.blade.php` but such file DOES NOT EXIST.

I think this issue probably needs to be addressed eventually.

## Summary by CodeRabbit

* **Chores**
  * Replaced Google reCAPTCHA with Cloudflare Turnstile across authentication flows; updated environment and service configuration to use Turnstile keys.
  * Updated dependencies and build config to support Turnstile and newer toolchain; bumped Node version.
* **Other**
  * Removed legacy client-side reCAPTCHA scripts and migrated UI components and styles to the Turnstile widget.


matiasperrone-exo:
### Reviewed Dev Notes

| # | Dev Note | Status |
|------|----------|--------|
| 1 | Package auto-discovers, NOT in config/app.php | PASS |
| 2 | siteKey (camelCase) + onSuccess in all React components | PASS |
| 3 | Controllers validate cf-turnstile-response (auto-rendered by widget) | PASS |
| 4 | login.js has two <Turnstile> instances (lines 187 + 266) | PASS |
| 5 | ForgotPasswordController now validates cf-turnstile-response | PASS |
| 6 | forgot_password.module.scss has .turnstile with 10px margins + center | PASS |
| 7 | 5 dead jQuery files deleted from public/assets/js/auth/ | PASS |
| 8 | Old RECAPTCHA_* env vars removed, new TURNSTILE_* added | PASS |

package.json
```diff
@@ -45,10 +44,10 @@
     "optimize-css-assets-webpack-plugin": "^6.0.1",
     "path": "^0.12.7",
     "postcss-loader": "^6.2.1",
-    "react": "^16.6.3",
+    "react": "^17.0.2",
     "react-bootstrap": "^0.33.1",
     "react-datetime": "^2.15.0",
-    "react-dom": "^16.4.1",
+    "react-dom": "^17.0.2",
```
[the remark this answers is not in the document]
matiasperrone-exo: done

resources/js/forgot_password/forgot_password.module.scss
```diff
@@ -14,6 +14,12 @@
     }
   }
 
+.turnstile {
+  margin-top: 10px;
+  margin-bottom: 10px;
+  text-align: center;
+}
```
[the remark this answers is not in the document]
matiasperrone-exo: done

.env.example
```diff
@@ -58,8 +58,8 @@ LOG_LEVEL=info
 
 EVENTBRITE_OAUTH2_PERSONAL_TOKEN=
 
-RECAPTCHA_PUBLIC_KEY=
-RECAPTCHA_PRIVATE_KEY=
+TURNSTILE_SITE_KEY=
+TURNSTILE_SECRET_KEY=
```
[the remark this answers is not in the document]
matiasperrone-exo: wontfix

smarcet:
@matiasperrone-exo u published secret keys to a public repository
image
please do check PR description

matiasperrone-exo:
> @matiasperrone-exo u published secret keys to a public repository image please do check PR description

Fixed

package.json
```diff
@@ -31,7 +31,6 @@
     "css-loader": "^6.7.1",
     "dotenv-webpack": "^1.7.0",
     "file-loader": "^6.2.0",
-    "font-awesome": "4.7.0",
     "history": "^4.7.2",
     "html-webpack-plugin": "^3.2.0",
```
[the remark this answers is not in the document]
matiasperrone-exo: This issue is bogus, `fontawesome` is present in the package.json line 90 (dependencies).

resources/js/login/login.js
```diff
@@ -263,10 +264,11 @@ const OTPInputForm = ({
...
                 {shouldShowCaptcha() &&
-                    <ReCAPTCHA
-                        className={styles.recaptcha}
-                        sitekey={captchaPublicKey}
-                        onChange={onChangeRecaptcha}
+                    <Turnstile
+                        className={styles.turnstile}
+                        siteKey={captchaPublicKey}
+                        options={{ responseFieldName: "cf-turnstile-response" }}
+                        onSuccess={onChangeCaptchaProvider}
```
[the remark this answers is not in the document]
matiasperrone-exo: fixed

webpack.common.js
```diff
@@ -83,6 +83,13 @@ module.exports = {
     ],
     module: {
         rules: [
+            {
+                // Fix for webpack 5 + ESM packages (e.g. @marsidev/react-turnstile) that import
+                // react/jsx-runtime without the .js extension, which webpack 5 requires in strict ESM mode.
+                test: /\.m?js/,
+                include: /node_modules/,
+                resolve: { fullySpecified: false }
```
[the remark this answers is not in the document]
matiasperrone-exo: fixed

.env.example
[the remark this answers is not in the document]
matiasperrone-exo: done

resources/views/auth/register.blade.php
```diff
@@ -58,7 +58,7 @@
             signInAction:'{{ URL::action("UserController@getLogin") }}',
             signUpAction: '{{ URL::action("Auth\RegisterController@register") }}',
             signUpError: signUpError,
-            captchaPublicKey: '{{ Config::get("recaptcha.public_key") }}',
+            captchaPublicKey: '{{ Config::get("services.turnstile.key") }}',
```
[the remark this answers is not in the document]
matiasperrone-exo: fixed

resources/js/login/login.js
```diff
@@ -184,10 +184,11 @@ const PasswordInputForm = ({
...
             {shouldShowCaptcha() && captchaPublicKey &&
-                <ReCAPTCHA
-                    className={styles.recaptcha}
-                    sitekey={captchaPublicKey}
-                    onChange={onChangeRecaptcha}
+                <Turnstile
+                    className={styles.turnstile}
+                    siteKey={captchaPublicKey}
+                    options={{ responseFieldName: "cf-turnstile-response" }}
+                    onSuccess={onChangeCaptchaProvider}
```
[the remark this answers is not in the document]
matiasperrone-exo: @smarcet check this one out.

app/Http/Controllers/UserController.php
```diff
@@ -420,7 +420,7 @@ public function postLogin()
             ];
 
             if ($login_attempts >= $max_login_attempts_2_show_captcha) {
-                $rules['g-recaptcha-response'] = 'required|recaptcha';
+                $rules['cf-turnstile-response'] = 'required|turnstile';
```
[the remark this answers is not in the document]
matiasperrone-exo: @smarcet check this one out.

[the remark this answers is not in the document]
smarcet: @matiasperrone-exo yes lets fix it many thanks

resources/js/login/login.js
```diff
@@ -262,11 +263,12 @@ const OTPInputForm = ({
                 <input type="hidden" value="otp" id="flow" name="flow"/>
                 <input type="hidden" value={otpCode} id="password" name="password"/>
                 <input type="hidden" value="email" id="connection" name="connection"/>
-                {shouldShowCaptcha() &&
-                    <ReCAPTCHA
-                        className={styles.recaptcha}
-                        sitekey={captchaPublicKey}
-                        onChange={onChangeRecaptcha}
+                {shouldShowCaptcha() && captchaPublicKey &&
```
martinquiroga-exo, line 275: OTP uses a separate form subtree from the password flow, while backend now requires `cf-turnstile-response` after the login-attempt threshold.

Please verify that the OTP submit path always includes a fresh Turnstile token after transitioning from password -> OTP, otherwise users who crossed the threshold in step 1 may be blocked in step 2 with a server-side captcha validation failure.

resources/js/login/login.js
martinquiroga-exo, line 194: Turnstile only auto-injects its hidden field [`captcha_value`] if the widget DOM stays mounted and the form submission uses the same DOM subtree.

This component has multiple state transitions (password --> OTP, login attempts, re renders), hence the token can be lost if the widget remounts before submit. 

Verify if we need to add a hidden input binding `captcha_value` into the <form>

martinquiroga-exo (changes requested):
@matiasperrone-exo please review changes

resources/js/login/login.js
matiasperrone-exo: we don't need to do that, no such thing happens.

resources/js/login/login.js
matiasperrone-exo: This is handled out of the box by Turnstile, not customization is required.

matiasperrone-exo: If you check the Turnstile library source, it calls `window.turnstile.render(containerDiv, params)`, so the widget renders entirely inside its own container div, and the hidden `<input name="cf-turnstile-response">` is injected within that container. Since the container div is inside each form's JSX tree, each form owns its own field. There is no cross-contamination.

So the transition works correctly:
  - Password form mounts, then its Turnstile renders into its `<form>`, the hidden field lives inside that form
  - User switches to OTP, then password form unmounts (taking its field with it), then OTP form mounts with a fresh Turnstile instance, then that widget renders its own hidden field inside the OTP <form>, so the submitted token is fresh

The `captcha_value` state being stale is a minor inconsistency in the React-side guard (`onAuthenticate` line 551), but it doesn't affect what actually gets submitted to the server, that comes from the DOM. And in practice, the user has to wait for an email before entering the OTP, giving the widget ample time to complete.

I add the state reset to be sure in here:
> ```
>    componentDidUpdate(prevProps, prevState) {
>       if (this.state.user_verified && this.existingUserCanContinue() && prevState.authFlow !== this.state.authFlow) {
>           this.setState({
>                ...this.state,
>                captcha_value: '',
>            });
>       }
>    }
> ```

app/Http/Controllers/UserController.php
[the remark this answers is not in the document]
matiasperrone-exo: Fixed, now the login attempts are retrieved from the DB instead of trusting the UI value.

resources/js/forgot_password/forgot_password.js
```diff
@@ -11,8 +11,8 @@ import Grid from "@material-ui/core/Grid";
 import TextField from "@material-ui/core/TextField";
 import Typography from "@material-ui/core/Typography";
 import Swal from "sweetalert2";
-import {MuiThemeProvider, createTheme} from "@material-ui/core/styles";
-import {useFormik} from "formik";
+import { MuiThemeProvider, createTheme } from "@material-ui/core/styles";
+import { useFormik } from "formik";
```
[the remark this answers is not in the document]
matiasperrone-exo: unrelated issue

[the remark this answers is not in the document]
matiasperrone-exo: Unrelated issue

matiasperrone-exo:
@martinquiroga-exo please review again.

app/Http/Controllers/UserController.php
```diff
@@ -410,7 +411,13 @@ public function postLogin()
             if (isset($data['password']))
                 $data['password'] = trim($data['password']);
 
-            $login_attempts = intval(Request::input('login_attempts'));
+            if (isset($data['username'])) {
+                $user = $this->auth_service->getUserByUsername($data['username']);
```
martinquiroga-exo: A few lines down we do `if ($login_attempts >= $max_login_attempts_2_show_captcha) {` so here if `username` is absent or `getUserByUsername()` returns `null` `$login_attempts` is never initialised before use

resources/js/login/login.js
```diff
@@ -183,11 +183,13 @@ const PasswordInputForm = ({
             <input type="hidden" value={userNameValue} id="username" name="username"/>
             <input type="hidden" value={csrfToken} id="_token" name="_token"/>
             <input type="hidden" value="password" id="flow" name="flow"/>
+            <input type="hidden" value={loginAttempts ?? 0} name="login_attempts"/>
```
martinquiroga-exo: This creates split-brain behaviour. It ties to the other comment from the issue in which `$login_attempts `is never initialised, in this scenario we might fallback to this and we will be trusting client side provided login attempts.

Why do we still need this if we are keeping count of attempts, as we should, server side?

martinquiroga-exo (changes requested):
@matiasperrone-exo please see comments.

Also, tests are missing, please include them for the following at least:

- Validation failure when cf-turnstile-response is missing.
- Login flow behaviour before/after CAPTCHA threshold.
- Server-side login-attempt lookup behaviour when user exists vs does not exist.
- Rendering of Turnstile on the thresholded login screen.
- Auth form submission when Turnstile is expired or not solved.

app/Http/Controllers/UserController.php
matiasperrone-exo: This statement is not true, the $login_attempts variable is initialized in line 399.

resources/js/login/login.js
matiasperrone-exo: Removed the hidden login_attempts inpunt element.

matiasperrone-exo:
Executing `docker compose exec app ./vendor/bin/phpunit tests/UserLoginTurnstileTest.php` locally works:
image

But it is failing in GitHub:
image

matiasperrone-exo:
On recreating the testing environment also works:

> $ `docker compose exec app vendor/bin/phpunit --log-junit results.xml`
> ```bash
> PHPUnit 11.5.21 by Sebastian Bergmann and contributors.
> 
> Runtime:       PHP 8.3.30
> Configuration: /var/www/phpunit.xml
> 
> ...............................................................  63 / 150 ( 42%)
> ............................................................... 126 / 150 ( 84%)
> ........................                                        150 / 150 (100%)
> 
> Time: 02:05.419, Memory: 141.00 MB
> 
> OK, but there were issues!
> Tests: 150, Assertions: 526, PHPUnit Deprecations: 1.
> ```

webpack.common.js
```diff
@@ -83,6 +83,13 @@ module.exports = {
     ],
     module: {
         rules: [
+            {
+                // Fix for webpack 5 + ESM packages (e.g. @marsidev/react-turnstile) that import
+                // react/jsx-runtime without the .js extension, which webpack 5 requires in strict ESM mode.
+                test: /\.m?js/,
```
romanetar: @matiasperrone-exo I think this should be "test: /\.m?js$/," to avoid the regex matches any path containing .js or .mjs anywhere (e.g., a directory named myjs/)

phpunit.xml
```diff
@@ -25,5 +25,7 @@
   </testsuites>
   <php>
     <env name="APP_ENV" value="testing"/>
+    <env name="TEST_USER_EMAIL" value="sebastian@tipit.net"/>
+    <env name="TEST_USER_PASSWORD" value="1Qaz2wsx!"/>
```
romanetar: @matiasperrone-exo please don't commit passwords to version control, even for test environments

romanetar (dismissed):
@matiasperrone-exo please review

phpunit.xml
matiasperrone-exo: This is in the actual seeding of the test code.

matiasperrone-exo: Here: https://github.com/OpenStackweb/openstackid/blob/main/database/seeds/TestSeeder.php#L251

webpack.common.js
matiasperrone-exo: Hello @romanetar! Thank you for the comment.

- Do you mean adding the "$" at the end?

The regular expression as-is is correct, let me explain a bit:
- This affects only files because is for "import" statements.
- \. means dot the "."character.
- m? means that the m may or may not be present.

Is used for adding to the file being imported, for instance `import Component from "/myjs"` may be  importing "myjs.js", "myjs.mjs", or if "my.js" is a directory then "my.js/index.js" or "myjs/index.mjs". 
Also this is only for tests because we do not add this tries to import by adding ".jsx" as default.

phpunit.xml
matiasperrone-exo: @smarcet what we should do?

app/Http/Controllers/UserController.php
```diff
@@ -410,7 +411,13 @@ public function postLogin()
             if (isset($data['password']))
                 $data['password'] = trim($data['password']);
 
-            $login_attempts = intval(Request::input('login_attempts'));
+            if (isset($data['username'])) {
```
smarcet: @matiasperrone-exo please lets rollback this change bc introduce another attack vector we will take care of this on another PR

smarcet (changes requested):
@matiasperrone please review comments

martinquiroga-exo (approved):
LGTM

phpunit.xml
matiasperrone-exo: moved to secrets

app/Http/Controllers/UserController.php
matiasperrone-exo: done

Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM php:8.3-fpm
 
 ARG DEBIAN_FRONTEND=noninteractive
 ARG NVM_VERSION="v0.40.3"
-ENV NODE_VERSION="18.20.4"
+ENV NODE_VERSION="20.20.2"
```
smarcet: @matiasperrone-exo this is breaking local build
node-sass should be replaced by sass-loader@12+

  - node-sass 8.x was the last line shipping prebuilt binaries for Node 14/16/18. There is no prebuilt binary for Node 20 or 22, so on Node 22 yarn falls back to compiling via node-gyp against an 
  unsupported nan/V8 ABI and fails (LibSass itself is also archived/EOL).                                                                                                                           
  - Even if compilation went through, LibSass (the C++ engine node-sass wraps) was retired in 2020 — it doesn't know modern Node toolchains.  

this PR fix this
https://github.com/OpenStackweb/openstackid/pull/131
please review and merge it to this branch
many thanks

smarcet (changes requested):
@matiasperrone-exo please review comments

Dockerfile
matiasperrone-exo: Done.

composer.json
```diff
@@ -49,6 +42,7 @@
     "php-opencloud/openstack": "3.10.0",
     "phpseclib/phpseclib": "^3.0.43",
     "predis/predis": "v2.2.2",
+    "ryangjchandler/laravel-cloudflare-turnstile": "^3.0",
```
caseylocker: **Blocker — vendor `Client::siteverify` has inverted logic at this version.**

`^3.0` resolves to v3.0.3. Upstream `src/Client.php` at that tag:

```php
if (! $response->ok()) {
    return SiteverifyResponse::success();        // HTTP non-2xx → pass validation (fail-open)
}
return SiteverifyResponse::failure($response->json('error-codes'));  // HTTP 2xx → always failure
```

Combined with `SiteverifyResponse::failure(array $errorCodes = [])` and the rule body `if ($response->success) return;`:

- **Valid token (`{success:true}` from Cloudflare):** `$response->json('error-codes')` is `null` → `failure(null)` violates the `array` typed param → TypeError → propagates out of the rule, through the validator, into each controller's outer `catch (Exception)` → generic error view or redirect. **Real users with valid tokens cannot complete registration / password reset / password set / email-verification resend / threshold-gated login.** (Confirmed null→array TypeError on local PHP 8.3.25; no `strict_types` needed.)
- **HTTP/network failure:** rule passes silently — fail-open on Cloudflare outage.
- **Invalid/expired token:** fails correctly.

v2.1.0's `TurnstileClient::siteverify` was correct (`success: $response->json('success')`); the inversion was introduced in the v3.0 rewrite. The PHPUnit suite doesn't catch this because `Turnstile::fake()` swaps in `FakeClient`, which is correct — but production traffic hits the real `Client.php`.

**Fix options (pick one):**
1. Pin `"ryangjchandler/laravel-cloudflare-turnstile": "^2.1"` and confirm Laravel compatibility.
2. File the upstream issue and re-pin to a 3.0.4 once cut.
3. Override the package's `ClientInterface` binding with a vendored correct `Client` in a local service provider. Surface area is tiny (`siteverify(string): SiteverifyResponse`).

Whichever path, please add one real-`Client` smoke test against Cloudflare's published always-pass / always-fail test keys (`1x...AA` / `2x...AB`) so this class of bug can't ship again silently.

app/Http/Controllers/UserController.php
```diff
@@ -420,7 +421,7 @@ public function postLogin()
             ];
 
             if ($login_attempts >= $max_login_attempts_2_show_captcha) {
-                $rules['g-recaptcha-response'] = 'required|recaptcha';
+                $rules['cf-turnstile-response'] = ['required', new Turnstile()];
```
caseylocker, line 424: **Blocker — this gate is effectively bypassed.**

The `$login_attempts` value used to attach this rule comes from `Request::input('login_attempts')` (line 414), but commit `3f48186b` removed the hidden `<input name="login_attempts">` from both the password and OTP forms in `resources/js/login/login.js` (lines 183-185 / 262-266). The DB-based lookup at lines 476-486 only runs **after** `$validator->passes()`, so it can't influence the rule.

Net effect: client never sends `login_attempts` → server always sees 0 → this `if` branch is never entered → `cf-turnstile-response` is not required → an attacker can POST without the captcha field after any number of failed attempts. The Turnstile widget still renders client-side once the React state exceeds the threshold, but server-side enforcement is no-op. Default threshold is 3 (`ServerConfigurationService::DefaultMaxFailedLoginAttempts2ShowCaptcha`).

@smarcet has already flagged this and asked to defer the proper fix (server-side session counter) to a separate PR — totally reasonable, but please surface it explicitly in the PR description as a known follow-up so it doesn't get lost. With this gate broken, `MaxFailed.Login.Attempts` account-lockout still protects single accounts, but distributed credential-stuffing across many usernames is no longer rate-limited at the IP/session level.

webpack.common.js
```diff
@@ -83,6 +84,13 @@ module.exports = {
...
```
caseylocker: Regex is unanchored — also matches `.jsx`, `.js.map`, and any path containing `"js/"` (e.g. `node_modules/myjs/foo.json`). @romanetar's suggestion of `/\.m?js$/` is correct; the rebuttal in the thread didn't address the anchoring issue. Tighten the trailing anchor:

```js
test: /\.m?js$/,
```

tests/UserLoginTurnstileTest.php
```diff
@@ -0,0 +1,223 @@
...
+
+use Auth\User;
+use Illuminate\Support\Facades\Session;
+use LaravelDoctrine\ORM\Facades\EntityManager;
+use RyanChandler\LaravelCloudflareTurnstile\Facades\Turnstile;
+
+/**
+ * Class UserLoginTurnstileTest
+ *
+ * Covers Cloudflare Turnstile integration in UserController::postLogin():
```
caseylocker, line 24: Class docstring is stale — it describes the reverted "DB user lookup" behavior. After commit `3f48186b` reverted `UserController::postLogin` back to `intval(Request::input('login_attempts'))`, `testLoginAttemptsDefaultToZeroForUnknownUsername` at line 146 became tautological: no `login_attempts` posted → 0 → no rule attached, regardless of whether the user exists in the DB. Update the docstring (and consider deleting the now-tautological test, or rewriting it to actually exercise the request-body counter behaviour).

tests/UserLoginTurnstileTest.php
```diff
@@ -0,0 +1,223 @@
...
+    private const CAPTCHA_THRESHOLD        = 3;
+
+    private string $testEmail;
+    private string $testPassword;
+
+    protected function prepareForTests(): void
+    {
+        parent::prepareForTests();
+        $this->testEmail    = env('TEST_USER_EMAIL');
+        $this->testPassword = env('TEST_USER_PASSWORD');
```
caseylocker, line 45: Two issues with the test fixture:

1. `env('TEST_USER_EMAIL')` / `env('TEST_USER_PASSWORD')` are read with no fallback and no `markTestSkipped`. If a developer's `.env` doesn't have these, several tests silently misclassify (only `testLoginAttemptsDefaultToZeroForUnknownUsername` is correct in that case). Either default to the seeded `sebastian@tipit.net` / `1Qaz2wsx!` from `TestSeeder.php:248-266`, or `markTestSkipped` when the env vars are missing.

2. Only login is covered by this file. The 5 other Turnstile-protected validators (`RegisterController:181`, `ForgotPasswordController:166`, `ResetPasswordController:117`, `PasswordSetController:159`, `EmailVerificationController:91`) have zero test coverage — and given Blocker 1 those are exactly the ones that would silently break in production. Adding minimal happy-path tests with `Turnstile::fake()` for each is cheap insurance.

resources/js/login/login.js
```diff
@@ -184,10 +184,11 @@ const PasswordInputForm = ({
             <input type="hidden" value={csrfToken} id="_token" name="_token"/>
             <input type="hidden" value="password" id="flow" name="flow"/>
             {shouldShowCaptcha() && captchaPublicKey &&
-                <ReCAPTCHA
-                    className={styles.recaptcha}
-                    sitekey={captchaPublicKey}
-                    onChange={onChangeRecaptcha}
+                <Turnstile
```
caseylocker, line 190: Only `onSuccess` is wired here (and in all five other widgets: `login.js:268`, `signup.js:278`, `forgot_password.js:140`, `reset_password.js:175`, `set_password.js:284`, `email_verification.js:124`). `@marsidev/react-turnstile` also exposes `onExpire` and `onError`.

Cloudflare's widget defaults to `refresh-expired: "auto"` so the **submitted** token is automatically refreshed — this is **not** a security issue. Two cosmetic gaps remain:

1. React `state.captcha_value` set on first `onSuccess` goes stale on token expiry. The client-side guard at line 552 (`if (this.state.captcha_value == '' && this.shouldShowCaptcha())`) won't re-block after expiry.
2. Cloudflare client-side errors / outages produce a silently-stuck widget with no UX feedback.

Low priority — wire `onExpire` (clear `captcha_value`) and `onError` (surface a message) at all six widgets if you want defensive UX.

caseylocker (changes requested):
Two blockers I'd want resolved before merge, plus a handful of mediums/lows posted inline.

**Blocker 1 — vendor `ryangjchandler/laravel-cloudflare-turnstile` v3.0.3 has inverted siteverify logic.** `src/Client.php` at the pinned commit `7da8024` returns `SiteverifyResponse::success()` on HTTP non-2xx (fail-open) and `SiteverifyResponse::failure($response->json('error-codes'))` on HTTP 2xx — which passes `null` to an `array`-typed param when Cloudflare returns `{success:true}`, raising TypeError. Real users with valid tokens cannot complete any captcha-protected form. Tests don't catch it because they bind `FakeClient` via `Turnstile::fake()`. v2.1.0 had correct logic; the v3 rewrite regressed it. See inline thread on `composer.json`.

**Blocker 2 — server-side captcha gate on login is effectively bypassed.** `login.js` no longer posts `login_attempts`, but `UserController::postLogin` still reads it from `Request::input()`. Net: server always sees 0, threshold rule never attaches, POST without `cf-turnstile-response` succeeds. Smarcet asked to defer the proper fix to a separate PR — flagging here so it doesn't get lost; please surface it in the PR description as a known follow-up. See inline thread on `UserController.php:424`.

**Other items posted inline:** unanchored webpack regex (`webpack.common.js:90`), test env-var fallback + coverage gap for the 5 other captcha-protected controllers, missing `onExpire`/`onError` on every Turnstile widget, stale class docstring on the test file.

**One item I can't post inline because the file isn't in the diff:** `resources/views/auth/email_verification_success.blade.php:6` still references `assets/js/auth/email-verification-complete.js` which doesn't exist. Either drop the `script_to(...)` line in this PR or carry it to a follow-up.

**What I verified as not concerning:** mechanical migration is clean across all 6 flows (controllers, blade templates, SCSS, JS); no residual `recaptcha` / `g-recaptcha` / `RECAPTCHA_*` references; `componentDidUpdate` correctly clears `captcha_value` on password→OTP transitions (addresses the earlier widget-remount concern); CSRF tokens preserved; the `'cf-turnstile-response' => ['required', new Turnstile()]` form is the vendor's documented usage. Node 22 + sass-loader@12 + webpack-cli@5.1.4 + `util:false` fallback all look correct.

webpack.common.js
matiasperrone-exo: done

tests/UserLoginTurnstileTest.php
matiasperrone-exo: 1. done
2. done

composer.json
matiasperrone-exo: done. Resolved with option 3

app/Http/Controllers/UserController.php
matiasperrone-exo: Done

resources/js/login/login.js
matiasperrone-exo: done

app/Services/TurnstileClient.php
```diff
@@ -0,0 +1,53 @@
...
+
+        if (! $response->ok()) {
+            return SiteverifyResponse::failure(['http-error']);
+        }
+
+        if ($response->json('success') === true) {
+            return SiteverifyResponse::success();
+        }
+
+        return SiteverifyResponse::failure($response->json('error-codes') ?? []);
```
caseylocker: **Security bypass — empty `error-codes` passes validation**

The vendor rule (`vendor/ryangjchandler/laravel-cloudflare-turnstile/src/Rules/Turnstile.php:21-32`) only calls `$fail()` inside a `foreach ($response->errorCodes as $errorCode)`. An empty array makes the loop a no-op, so the validator records **no failure** even though `success === false`.

This line returns `failure([])` whenever Cloudflare responds 200 with `{"success": false}` and an absent or empty `error-codes` field. A malformed Cloudflare response or future API drift would silently let invalid tokens through.

Required fix:

```php
return SiteverifyResponse::failure($response->json('error-codes') ?: ['internal-error']);
```

(`?:` instead of `??` so an empty array also falls back, not just `null`.)

app/Http/Controllers/UserController.php
caseylocker, line 424: **Login captcha gate trusts a client-controlled counter**

At line 414 (just above this hunk), `$login_attempts = intval(Request::input('login_attempts'))` reads the failed-attempt count from the POST body. An attacker who omits the field (or posts `0`) bypasses this Turnstile rule entirely — and `tests/UserLoginTurnstileTest.php:150-162` (`testOmittedLoginAttemptsFieldDefaultsToZeroNoCaptchaRequired`) documents that exact behavior as expected.

This was pre-existing reCAPTCHA tech debt and there's a follow-up ticket ([86b9v1acw](https://app.clickup.com/t/86b9v1acw)) for the server-side fix. But the PR description claims this migration "hardens" the gate, and the new server-side Turnstile coverage isn't real coverage as long as the gate is forgeable.

Two options to unblock:

1. **Preferred** — land the server-side counter from 86b9v1acw in this PR so the migration genuinely delivers the hardening it claims, **or**
2. **Acceptable** — drop the "hardening" language from the PR description, add a code comment at line 414 pointing at 86b9v1acw, and confirm with @smarcet that we're knowingly merging a captcha that an attacker can bypass.

Either way, AC5 ("Login form still conditionally shows the captcha widget only after N failed login attempts") is satisfied at the UI layer, not at the validation layer.

tests/UserLoginTurnstileTest.php
```diff
@@ -0,0 +1,244 @@
...
+
+    public function testLoginAtThresholdWithValidTokenPassesValidation(): void
+    {
+        $user = $this->getTestUser();
+
+        $this->fakeTurnstilePass();
+
+        $this->postLogin([
+            'cf-turnstile-response' => 'dummy-token-accepted-by-mock',
+            'login_attempts' => 1
```
caseylocker: **This test doesn't exercise the threshold path**

`CAPTCHA_THRESHOLD` is `3` (line 35) but this test posts `login_attempts => 1`, which is below threshold — so `UserController.php:423` never adds the Turnstile rule, and the assertion on line 144 passes trivially because the rule wasn't applied at all.

The test's name claims it's the happy-path coverage for "valid token clears the captcha rule," but right now it's just a duplicate of `testLoginBelowThresholdDoesNotRequireTurnstile`. Please change line 141 to:

```php
'login_attempts' => self::CAPTCHA_THRESHOLD
```

This is the only test in the file that actually verifies a valid token *clears* the rule under enforcement, so without this fix AC4 has no positive-path coverage on the login form.

caseylocker (changes requested):
Migration is structurally complete and all six ACs map to concrete code, but a few issues need to be resolved before merge. Independent second-pass review corroborated the findings below.

**Blocking (inline comments below):**
1. `app/Services/TurnstileClient.php:46` — empty `error-codes` from Cloudflare silently passes the vendor rule (`$fail()` is never called). Security bypass with a one-line fix.
2. `app/Http/Controllers/UserController.php:414/424` — login captcha gate is forgeable via the request-body `login_attempts` counter.
3. `tests/UserLoginTurnstileTest.php:141` — the "valid token clears rule" test posts `login_attempts => 1` against a threshold of `3`, so it never exercises the rule it claims to cover (this is the only positive-path test for AC4 on the login form).

**Blocking (no good inline target — autogenerated file):**
4. `composer.lock` is out of date with `composer.json`. Reproduced locally:
```
$ composer validate --strict --no-check-publish
# Lock file errors
- The lock file is not up to date with the latest changes in composer.json,
  it is recommended that you run `composer update` or `composer update <package name>`.
```
CI may be green against a stale lock but clean checkouts will diverge. Please run `composer update ryangjchandler/laravel-cloudflare-turnstile` (or `composer update --lock` if only metadata drifted) and recommit `composer.lock`.

**Strongly recommended but non-blocking** (happy to take as follow-up commits or separate tickets):
- Add `idempotency_key` (UUID per call) to the siteverify request in `TurnstileClient::siteverify()` — `Http::retry(3, 100)` will otherwise consume tokens twice on transient failures and produce `timeout-or-duplicate` rejections for legitimate users on the second attempt. Cloudflare documents this for exactly this case: https://developers.cloudflare.com/turnstile/get-started/server-side-validation/
- Catch `Illuminate\Http\Client\ConnectionException` inside `TurnstileClient::siteverify()` and return `SiteverifyResponse::failure(['internal-error'])` so network errors surface as a clean validation error instead of the generic auth error view.
- Also pass `remoteip` to siteverify (optional but recommended hardening — binds the verification to the originating client IP).
- Tighten the composer constraint to `~3.0.3` (or pin to the affected version) so a future upstream fix in `ryangjchandler/laravel-cloudflare-turnstile` doesn't silently get shadowed by our override. Linking the upstream bug in the override's docblock would also help future maintainers.
- `tests/UserLoginTurnstileTest::prepareForTests` calls `markTestSkipped()` when `TEST_USER_EMAIL`/`TEST_USER_PASSWORD` are absent. The CI workflows pull from `secrets.TEST_USER_EMAIL` / `secrets.TEST_USER_PASSWORD` — please confirm those secrets exist on the repo, otherwise the entire login-captcha suite can silently no-op in CI. Better long-term: create the test user via factory/fixture so the suite is self-contained.
- Out-of-scope toolchain bumps (Node 18→22, React 16→17, react-redux 5→7, `node-sass`→`sass`, webpack-cli 4→5, several removed packages) are riding along with this PR. Please confirm non-auth pages were smoke-tested under React 17 / Dart Sass, or split the toolchain upgrade into its own PR — the new tests only cover auth flows so regressions elsewhere wouldn't be caught.

**Note on AC1/AC5/AC6:**
- AC1 grep is clean (no remaining `recaptcha`/`g-recaptcha`/`ReCAPTCHA` refs in production code) ✓
- AC5 is satisfied at the UI conditional-render layer but not at the validation layer (see comment on UserController) — wording in the PR description's "hardening" claim may want a softer framing or pair with the fix for [86b9v1acw](https://app.clickup.com/t/86b9v1acw).
- AC6 ✓

Nice work on the custom `TurnstileClient` override and the dual `BrowserKitTesting` suite — those are the right shapes. The remaining issues are tractable.

tests/UserLoginTurnstileTest.php
matiasperrone-exo: done

app/Http/Controllers/UserController.php
matiasperrone-exo: Wont fix. 

A different PR will address this issue. Related task:
Ref: https://app.clickup.com/t/86b9v1acw

app/Services/TurnstileClient.php
matiasperrone-exo: done

tests/UserLoginTurnstileTest.php
matiasperrone-exo: done

caseylocker (approved):
If the yarn issues are in fact fixed and/or out of scope then approved. @smarcet please have a look.

smarcet (approved):
LGTM

smarcet (approved):
LGTM

PR #126, opened 2026-04-22
Feature | Implement Two-Factor Authentication (2FA) support for users

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b9h8xt1

## Depends on
- https://github.com/OpenStackweb/openstackid/pull/125 - https://app.clickup.com/t/86b9e8wm7

## Changes

 - Created PHPUnit test suit "Two Factor Authentication Test Suite" Modified app/libs/Auth/Models/User.php:
  - public const ValidMFAMethods = ['email_otp'];
  - 3 new Doctrine-mapped private fields (two_factor_enabled, two_factor_method, two_factor_enforced_at) matching the Phase 1 migration columns
  - Constructor initializers for the three fields
  - Getters/setters: isTwoFactorEnabled/setTwoFactorEnabled, getTwoFactorMethod/setTwoFactorMethod, getTwoFactorEnforcedAt/setTwoFactorEnforcedAt
  - shouldRequire2FA() — config-driven via two_factor.enforced_groups, falls through to the stored flag
  - enable2FA(string $method) — whitelists via ValidMFAMethods, throws ValidationException otherwise
  - getAvailableTwoFactorMethods() / isTwoFactorMethodEnable()
  - Phase II/III stubs returning false: isPhoneNumberVerified, isTOTPConfirmed, isPassKeyEnabled
  - Created config/two_factor.php with enforced_groups referencing IGroupSlugs constants (super-admins, administrators, oauth2-server-admins, openid-server-admins).
  - Created tests/unit/UserTwoFactorTest.php — 11 test methods (25 assertions), all green.
 
### Verification:
  - doctrine:schema:validate: no new diffs on users relating to the 2FA columns — only the pre-existing documented noise (signed/unsigned, index renames).
  - UserTwoFactorTest: 11/11 passing.
  - TwoFactorRepositoriesTest (Phase 1): 3/3 still passing.
  
  ----

#### Current state

The User entity (Auth\User) has no 2FA-related properties or methods. Role checking exists via isSuperAdmin() and isAdmin() but there is no mechanism to determine whether a user requires two-factor authentication.

#### Target state

The User entity exposes the three new database fields as Doctrine-mapped properties and provides the business logic methods needed by the controller and services:
shouldRequire2FA(), enable2FA(), getAvailableTwoFactorMethods(), isTwoFactorMethodEnable(), getTwoFactorMethod(), and the ValidMFAMethods constant.

### TASKS

  - Add Doctrine property mappings for two_factor_enabled, two_factor_method, two_factor_enforced_at to the User entity with getters and setters
  - Add the ValidMFAMethods constant array with value ['email_otp'] (sms_otp, totp, passkey are stubs for Phase II/III)
  - Implement shouldRequire2FA(): returns true if user isAdmin(), otherwise returns two_factor_enabled value
  - Implement enable2FA(string $method): validates method against ValidMFAMethods, sets two_factor_enabled=true, two_factor_method, and two_factor_enforced_at=now()
  - Implement getAvailableTwoFactorMethods(): returns ['email_otp'] if email is verified (Phase I always returns email_otp for enrolled users). Stubs for isPhoneNumberVerified(),
  isTOTPConfirmed(), isPassKeyEnabled() return false.
  - Implement isTwoFactorMethodEnable(string $method): returns whether the method is in getAvailableTwoFactorMethods()
  - Implement isEmailVerified() if not already present (check existing User entity for email_verified or active status)
  - Write unit tests for shouldRequire2FA() with: super-admin user, admin user, regular user with 2FA enabled, regular user with 2FA disabled
  - Write unit tests for enable2FA() with valid and invalid method values
  - Write unit tests for getAvailableTwoFactorMethods() and isTwoFactorMethodEnable()

### ACCEPTANCE CRITERIA

  - shouldRequire2FA() returns true for any user in super-admins or administrators groups regardless of two_factor_enabled value
  - shouldRequire2FA() returns the value of two_factor_enabled for non-admin users
  - enable2FA('email_otp') sets all three fields correctly; enable2FA('invalid') throws ValidationException
  - getAvailableTwoFactorMethods() returns ['email_otp'] for active users in Phase I
  - isTwoFactorMethodEnable('email_otp') returns true; isTwoFactorMethodEnable('sms_otp') returns false in Phase I
  - All unit tests pass

### DEVELOPMENT NOTES

  Key files:
  - Modified: app/libs/Auth/User.php (or Auth\User depending on namespace)
  - New: tests/Unit/UserTwoFactorTest.php

### Gotchas:
  - User::isAdmin() checks SuperAdminGroup OR AdminGroup. shouldRequire2FA() must use isAdmin() not just check one group.
  - The SDS specifies that Phase I enforced_groups config also includes oauth2-server-admins and openid-server-admins. shouldRequire2FA() should check membership in ALL groups listed in
  config('two_factor.enforced_groups'), not hardcode group slugs.
  - isEmailVerified() semantics: the existing User model may use 'active' status or a specific email_verified flag. Check the existing codebase.
  - Phase II/III stub methods (isPhoneNumberVerified, isTOTPConfirmed, isPassKeyEnabled) must return false so getAvailableTwoFactorMethods() only returns email_otp in Phase I.

*Out of scope:* Database migration (Ticket 1), controller integration, service layer.

### DIFF
--- a/app/libs/Auth/Models/User.php
+++ b/app/libs/Auth/Models/User.php
@@ -76,6 +76,15 @@ class User extends BaseEntity
         self::SpamTypeHam
     ];
 
+    public const MFAMethod_OTP = 'email_otp';
+    public const MFAMethod_SMS = 'sms_otp';
+    public const MFAMethod_TOTP = 'totp';
+    public const MFAMethod_PASSKEY = 'passkey';
+
+    public const ValidMFAMethods = [
+        self::MFAMethod_OTP
+    ];
+
     /**
      * @var string
      */
@@ -303,6 +312,25 @@ class User extends BaseEntity
      */
     #[ORM\Column(name: 'email_verified_date', nullable: true, type: 'datetime')]
     private $email_verified_date;
+
+    /**
+     * @var bool
+     */
+    #[ORM\Column(name: 'two_factor_enabled', type: 'boolean', options: ['default' => false])]
+    private $two_factor_enabled;
+
+    /**
+     * @var string
+     */
+    #[ORM\Column(name: 'two_factor_method', type: 'string', length: 32, options: ['default' => self::MFAMethod_OTP])]
+    private $two_factor_method;
+
+    /**
+     * @var \DateTime|null
+     */
+    #[ORM\Column(name: 'two_factor_enforced_at', nullable: true, type: 'datetime')]
+    private $two_factor_enforced_at;
+
     /**
      * @var string
      */
@@ -457,6 +485,9 @@ class User extends BaseEntity
         parent::__construct();
         $this->active = true;
         $this->email_verified = false;
+        $this->two_factor_enabled = false;
+        $this->two_factor_method = self::MFAMethod_OTP;
+        $this->two_factor_enforced_at = null;
         // user profile settings
         $this->public_profile_show_photo = false;
         $this->public_profile_show_email = false;
@@ -2357,6 +2388,144 @@ SQL;
     public function getAuthPasswordName()
     {
         return 'password';
+    }
+
+    // --- Two-factor authentication ---------------------------------------
+
+    public function isTwoFactorEnabled(): bool
+    {
+        return (bool) $this->two_factor_enabled;
+    }
+
+    public function setTwoFactorEnabled(bool $enabled): void
+    {
+        $this->two_factor_enabled = $enabled;
+    }
+
+    public function getTwoFactorMethod(): string
+    {
+        return $this->two_factor_method;
+    }
+
+    /**
+     * @throws ValidationException
+     */
+    protected function setTwoFactorMethod(string $method): void
+    {
+        $this->two_factor_method = $method;
+    }
+
+    public function getTwoFactorEnforcedAt(): ?\DateTime
+    {
+        return $this->two_factor_enforced_at;
+    }
+
+    public function setTwoFactorEnforcedAt(?\DateTime $at): void
+    {
+        $this->two_factor_enforced_at = $at;
+    }
+
+    /**
+     * Whether this user is required to complete 2FA to sign in.
+     *
+     * A user is required when they belong to any of the groups listed in
+     * config('two_factor.enforced_groups'); otherwise the stored flag applies.
+     */
+    public function shouldRequire2FA(): bool
+    {
+        $enforcedGroups = config('two_factor.enforced_groups', []);
+        foreach ($enforcedGroups as $slug) {
+            if($this->belongToGroup($slug)) {
+                return true;
+            }
+        }
+        return (bool) $this->two_factor_enabled;
+    }
+
+    /**
+     * @throws ValidationException
+     */
+    public function enable2FA(string $method): void
+    {
+        $availableMethods = $this->getAvailableTwoFactorMethods();
+        if(!in_array($method, self::ValidMFAMethods, true)) {
+            throw new ValidationException(
+                sprintf(
+                    "Invalid 2FA method '%s'. Allowed methods: %s. Enabled methods: %s",
+                    $method,
+                    implode(', ', self::ValidMFAMethods),
+                    implode(', ', $availableMethods)
+                )
+            );
+        }
+
+        if(!in_array($method, $availableMethods, true)) {
+            throw new ValidationException(
+                sprintf(
+                    "Disabled 2FA method '%s'. Enabled methods: %s",
+                    $method,
+                    implode(', ', $availableMethods)
+                )
+            );
+        }
+
+        $this->setTwoFactorMethod($method);
+        $this->setTwoFactorEnabled(true);
+        $this->setTwoFactorEnforcedAt(new \DateTime('now', new \DateTimeZone('UTC')));
+    }
+
+    public function disable2FA(): void
+    {
+        $this->setTwoFactorEnabled(false);
+        $this->setTwoFactorEnforcedAt(null);
+    }
+
+    /**
+     * Returns the set of 2FA methods currently available to this user.
+     * Phase I only supports email_otp; other methods are stubs that will
+     * light up in Phase II/III once the backing verifications exist.
+     *
+     * @return string[]
+     */
+    public function getAvailableTwoFactorMethods(): array
+    {
+        $methods = [];
+        if($this->isEmailVerified() && in_array(self::MFAMethod_OTP, self::ValidMFAMethods, true)) {
+            $methods[] = self::MFAMethod_OTP;
+        }
+        if($this->isPhoneNumberVerified() && in_array(self::MFAMethod_SMS, self::ValidMFAMethods, true)) {
+            $methods[] = self::MFAMethod_SMS;
+        }
+        if($this->isTOTPConfirmed() && in_array(self::MFAMethod_TOTP, self::ValidMFAMethods, true)) {
+            $methods[] = self::MFAMethod_TOTP;
+        }
+        if($this->isPassKeyEnabled() && in_array(self::MFAMethod_PASSKEY, self::ValidMFAMethods, true)) {
+            $methods[] = self::MFAMethod_PASSKEY;
+        }
+        return $methods;
+    }
+
+    public function isTwoFactorMethodEnabled(string $method): bool
+    {
+        return in_array($method, $this->getAvailableTwoFactorMethods(), true);
+    }
+
+    // Phase II stub
+    public function isPhoneNumberVerified(): bool
+    {
+        return false;
+    }
+
+    // Phase III stub
+    public function isTOTPConfirmed(): bool
+    {
+        return false;
+    }
+
+    // Phase III stub
+    public function isPassKeyEnabled(): bool
+    {
+        return false;
     }
 
 }
--- a/tests/unit/UserTwoFactorTest.php
+++ b/tests/unit/UserTwoFactorTest.php
@@ -1,0 +1,246 @@
+<?php
+namespace Tests\unit;
+
+/**
+ * Copyright 2026 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+use App\libs\Auth\Models\IGroupSlugs;
+use Auth\Group;
+use Auth\User;
+use Illuminate\Support\Facades\Config;
+use models\exceptions\ValidationException;
+use Tests\TestCase;
+
+/**
+ * Class UserTwoFactorTest
+ * Unit tests for the 2FA-related methods on the User entity.
+ */
+class UserTwoFactorTest extends TestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+        Config::set('two_factor.enforced_groups', [
+            IGroupSlugs::SuperAdminGroup,
+            IGroupSlugs::AdminGroup,
+            IGroupSlugs::OAuth2ServerAdminGroup,
+            IGroupSlugs::OpenIdServerAdminsGroup,
+        ]);
+    }
+
+    private function buildGroup(string $slug): Group
+    {
+        $group = new Group();
+        $group->setName($slug);
+        $group->setSlug($slug);
+        return $group;
+    }
+
+    private function assignGroups(User $user, array $groups): void
+    {
+        $reflection = new \ReflectionClass(User::class);
+        $property = $reflection->getProperty('groups');
+        $property->setAccessible(true);
+        $collection = $property->getValue($user);
+        foreach ($groups as $group) {
+            $collection->add($group);
+        }
+    }
+
+    private function setEmailVerified(User $user, bool $verified): void
+    {
+        $reflection = new \ReflectionClass(User::class);
+        $property = $reflection->getProperty('email_verified');
+        $property->setAccessible(true);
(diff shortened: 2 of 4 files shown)


app/libs/Auth/Models/User.php
```diff
@@ -2359,4 +2421,127 @@ public function getAuthPasswordName()
...
+
+    /**
+     * Whether this user is required to complete 2FA to sign in.
+     *
+     * A user is required when they belong to any of the groups listed in
+     * config('two_factor.enforced_groups'); otherwise the stored flag applies.
+     */
+    public function shouldRequire2FA(): bool
+    {
+        return ($this->isAdmin() || (bool) $this->two_factor_enabled);
```
martinquiroga-exo: This should check for members listed in `two_factor.enforced_groups` not only `isAdmin()`

```
    public function isAdmin(): bool
    {
        return $this->belongToGroup(IGroupSlugs::SuperAdminGroup) || $this->belongToGroup(IGroupSlugs::AdminGroup);
    }
```
It only checks for `SuperAdminGroup` and `AdminGroup` which is not enough:

```
    'enforced_groups' => [
        IGroupSlugs::SuperAdminGroup,
        IGroupSlugs::AdminGroup,
        IGroupSlugs::OAuth2ServerAdminGroup,
        IGroupSlugs::OpenIdServerAdminsGroup,
    ],
```

tests/unit/UserTwoFactorTest.php
martinquiroga-exo, line 1: All tests pass only if `shouldRequire2FA()` ignores `two_factor.enforced_groups`

tests/unit/UserTwoFactorTest.php
```diff
@@ -0,0 +1,216 @@
...
+    public function testShouldRequire2FA_adminUser(): void
+    {
+        $user = new User();
+        $this->assignGroups($user, [$this->buildGroup(IGroupSlugs::AdminGroup)]);
+
+        $this->assertFalse($user->isTwoFactorEnabled());
+        $this->assertTrue($user->shouldRequire2FA());
+    }
+
+    public function testShouldRequire2FA_oauth2AdminUser(): void
```
martinquiroga-exo: This proves OAuth2 admin is not covered

tests/unit/UserTwoFactorTest.php
```diff
@@ -0,0 +1,216 @@
...
+
+        $reflection = new \ReflectionClass(User::class);
+        $setter = $reflection->getMethod('setTwoFactorMethod');
+        $setter->setAccessible(true);
+
+        $this->expectException(ValidationException::class);
+        $setter->invoke($user, 'bogus');
+    }
+
+    public function testShouldRequire2FA_emptyEnforcedGroups_fallsThroughToFlag(): void
```
martinquiroga-exo, line 249: This test only validates the fallback path (`two_factor_enabled` flag) and does not actually prove that `config('two_factor.enforced_groups')` is used.
Even after fixing `shouldRequire2FA()` to read config, this test would still pass if config logic is broken again.
Add a complementary test where a user belongs to a group in `enforced_groups` and `two_factor_enabled = false`, asserting that `shouldRequire2FA()` returns true. This ensures config-driven enforcement is actually covered.

martinquiroga-exo (changes requested):
@matiasperrone-exo please see comments. Thanks.

app/libs/Auth/Models/User.php
matiasperrone-exo: fixed

tests/unit/UserTwoFactorTest.php
matiasperrone-exo: fixed

tests/unit/UserTwoFactorTest.php
matiasperrone-exo: fixed

tests/unit/UserTwoFactorTest.php
matiasperrone-exo: fixed

matiasperrone-exo:
@martinquiroga-exo please review again

martinquiroga-exo (approved):
LGTM

caseylocker:
@matiasperrone-exo — small rename request to align this PR with the SDS.

`sds/idp-mfa.md` uses `isTwoFactorMethodEnable()` (no trailing `d`) in 5 places: lines 438, 1988, 2064, 2151, 2186 — including the controller-flow pseudo-code at 438 and the AuthService snippets at 2151/2186 that the downstream MFA Challenge Strategy ticket (86b9j5jup) will reference.

No in-flight PR currently calls either spelling — PR #125 (base) and PR #127 (AuthService validateCredentials) don't reference it — so the rename is safe to make on this branch only:

- `app/libs/Auth/Models/User.php:2547` — rename `isTwoFactorMethodEnabled` → `isTwoFactorMethodEnable`
- `tests/unit/UserTwoFactorTest.php:190-194` — update the 5 assertion call sites (test method name `testIsTwoFactorMethodEnable` already matches the SDS spelling)

The current spelling is more grammatical, but updating the SDS in lockstep is more disruptive than matching it here.

app/libs/Auth/Models/User.php
```diff
@@ -2359,4 +2426,145 @@ public function getAuthPasswordName()
...
+    // --- Two-factor authentication ---------------------------------------
+
+    public function isTwoFactorEnabled(): bool
+    {
+        return (bool) $this->two_factor_enabled;
+    }
+
+    public function setTwoFactorEnabled(bool $enabled): void
+    {
+        $this->two_factor_enabled = $this->shouldRequire2FA() || $enabled;
```
caseylocker: This setter overlays policy onto the persisted flag: `shouldRequire2FA()` short-circuits to `true` via group membership for any user in `enforced_groups`, so `setTwoFactorEnabled(false)` silently persists `true` for those users.

Two consequences:
- An admin who is later demoted has a stale persisted `true` even though no caller intended it.
- Callers reasoning "I set it to false, so the column is false" are wrong for any enforced-group member.

Both the SDS § 6 and the ticket DOD prescribe direct assignment inside `enable2FA()` (`$this->two_factor_enabled = true;`) with no `setTwoFactorEnabled()` setter at all. Suggest dropping the policy coupling — have the setter just write the flag faithfully and let `shouldRequire2FA()` be the single source of policy on read:

```php
public function setTwoFactorEnabled(bool $enabled): void
{
    $this->two_factor_enabled = $enabled;
}
```

No test currently covers the `admin + setTwoFactorEnabled(false)` case.

app/libs/Auth/Models/User.php
```diff
@@ -2359,4 +2426,145 @@ public function getAuthPasswordName()
...
+                sprintf(
+                    "Disabled 2FA method '%s'. Enabled methods: %s",
+                    $method,
+                    implode(', ', $availableMethods)
+                )
+            );
+        }
+
+        $this->two_factor_method = $method;
+    }
```
caseylocker, line 2416: This setter is unreachable from any public caller. `enable2FA()` (User.php:2505-2520) already pre-validates `$method` against `getAvailableTwoFactorMethods()`, which is a strict subset of `ValidMFAMethods`, so a method that would fail the inner `ValidMFAMethods` check can never reach this setter through public code. The reflection-based test `testSetTwoFactorMethod_invalidMethod_throws` (`tests/unit/UserTwoFactorTest.php:197-209`) is the only thing exercising the path.

The SDS § 6 and the ticket DOD both define a single validation tier inside `enable2FA()` — no `setTwoFactorMethod()` setter is prescribed. Suggest either:
- Drop the setter entirely (single source of truth = `enable2FA()`'s pre-check) and remove the reflection test, or
- Inline the assignment inside `enable2FA()` directly (`$this->two_factor_method = $method;`) matching the SDS pattern at line 2024 of `sds/idp-mfa.md`.

Either way, the second-tier validation is gold-plating beyond spec.

app/Http/Controllers/SocialLoginController.php
```diff
@@ -157,7 +157,7 @@ public function callback($provider)
             if(!$user->canLogin()) {
                 throw new AuthenticationException
                 (
-                    "We are sorry, your username does not match an existing record."
+                    "username does not match an existing record."
```
caseylocker: This change is unrelated to the 2FA work in the rest of the PR. The ticket explicitly lists controller integration as out of scope:

> Out of scope: Database migration (Ticket 1), controller integration, service layer.

Recommend reverting this line and splitting it into its own PR. As a side note, the new copy reads as a bare fragment compared to the original ("We are sorry, your username does not match an existing record.") — if the goal is to shorten error copy, that's a separate UX discussion.

caseylocker (commented):
Three more findings on top of the rename request, ordered by severity. Each is also added as a line-level comment below.

1. **`setTwoFactorEnabled()` couples write-state to policy** — `app/libs/Auth/Models/User.php:2436-2438`. `setTwoFactorEnabled(false)` silently persists `true` for users in `enforced_groups` because `shouldRequire2FA()` short-circuits before reading the flag. Both the SDS § 6 and the ticket DOD prescribe direct assignment in `enable2FA()` with no policy-overlaying setter.

2. **Dead validation tier in `setTwoFactorMethod()`** — `app/libs/Auth/Models/User.php:2449-2473`. The inner `ValidMFAMethods` check is unreachable from public callers; only the reflection test exercises it. Both the SDS and the ticket prescribe single-tier validation in `enable2FA()` with no `setTwoFactorMethod()` setter at all.

3. **Unrelated change in `SocialLoginController.php:160`** — error message wording shortened. Out of scope by the ticket's own statement ("Out of scope: ... controller integration, service layer."). Recommend reverting and splitting to its own PR.

For context: I cross-checked all three against the SDS (`sds/idp-mfa.md` § 3.1, § 4.3, § 6) and the ClickUp ticket DOD (86b9h8xt1). All three are supported by both — they are additions/scope drift the spec does not ask for.

A second-pass review (Codex) independently confirmed all three. Two additional concerns I am not posting here:
- A minor design concern about `two_factor_method`'s non-null `email_otp` default — withdrawn after rechecking, since SDS § 3.1 explicitly prescribes that default.
- A defensive-guard suggestion in `shouldRequire2FA()` against misconfigured `enforced_groups` (typo'd slugs, non-array) — nice-to-have, not required.

caseylocker (changes requested):
@matiasperrone-exo please see comments. Thanks.

app/libs/Auth/Models/User.php
matiasperrone-exo: done

app/libs/Auth/Models/User.php
matiasperrone-exo: Moved validations to enable2FA()

app/libs/Auth/Models/User.php
matiasperrone-exo: done

app/libs/Auth/Models/User.php
matiasperrone-exo: Validations inside `enable2FA()` the setter in now clean

matiasperrone-exo:
> @matiasperrone-exo — small rename request to align this PR with the SDS.
> 
> `sds/idp-mfa.md` uses `isTwoFactorMethodEnable()` (no trailing `d`) in 5 places: lines 438, 1988, 2064, 2151, 2186 — including the controller-flow pseudo-code at 438 and the AuthService snippets at 2151/2186 that the downstream MFA Challenge Strategy ticket (86b9j5jup) will reference.
> 
> No in-flight PR currently calls either spelling — PR #125 (base) and PR #127 (AuthService validateCredentials) don't reference it — so the rename is safe to make on this branch only:
> 
> * `app/libs/Auth/Models/User.php:2547` — rename `isTwoFactorMethodEnabled` → `isTwoFactorMethodEnable`
> * `tests/unit/UserTwoFactorTest.php:190-194` — update the 5 assertion call sites (test method name `testIsTwoFactorMethodEnable` already matches the SDS spelling)
> 
> The current spelling is more grammatical, but updating the SDS in lockstep is more disruptive than matching it here.

We are going to correct this typo in the SDS

app/Http/Controllers/SocialLoginController.php
matiasperrone-exo: reverted

matiasperrone-exo:
@caseylocker please review it again, I addressed all the named issues by moving the validation away from the setters, and leave the setters clean also reverted the unexpected changes in SocialLoginController.

> Three more findings on top of the rename request, ordered by severity. Each is also added as a line-level comment below.
> 
> 1. **`setTwoFactorEnabled()` couples write-state to policy** — `app/libs/Auth/Models/User.php:2436-2438`. `setTwoFactorEnabled(false)` silently persists `true` for users in `enforced_groups` because `shouldRequire2FA()` short-circuits before reading the flag. Both the SDS § 6 and the ticket DOD prescribe direct assignment in `enable2FA()` with no policy-overlaying setter.
> 2. **Dead validation tier in `setTwoFactorMethod()`** — `app/libs/Auth/Models/User.php:2449-2473`. The inner `ValidMFAMethods` check is unreachable from public callers; only the reflection test exercises it. Both the SDS and the ticket prescribe single-tier validation in `enable2FA()` with no `setTwoFactorMethod()` setter at all.
> 3. **Unrelated change in `SocialLoginController.php:160`** — error message wording shortened. Out of scope by the ticket's own statement ("Out of scope: ... controller integration, service layer."). Recommend reverting and splitting to its own PR.
> 
> For context: I cross-checked all three against the SDS (`sds/idp-mfa.md` § 3.1, § 4.3, § 6) and the ClickUp ticket DOD (86b9h8xt1). All three are supported by both — they are additions/scope drift the spec does not ask for.
> 
> A second-pass review (Codex) independently confirmed all three. Two additional concerns I am not posting here:
> 
> * A minor design concern about `two_factor_method`'s non-null `email_otp` default — withdrawn after rechecking, since SDS § 3.1 explicitly prescribes that default.
> * A defensive-guard suggestion in `shouldRequire2FA()` against misconfigured `enforced_groups` (typo'd slugs, non-array) — nice-to-have, not required.

caseylocker (approved):
Based on the requirements/ticket this LGTM.

PR #127, opened 2026-04-27
Feature | Add AuthService validateCredentials method

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86b9j51aa

## Depends on:

- https://app.clickup.com/t/86b9h8xt1 - https://github.com/OpenStackweb/openstackid/pull/126

## Summary

Adds `validateCredentials()` and `loginUser()` to `AuthService` as the foundation for the MFA two-step login flow. The new methods split what was previously a single `login()` call into two discrete phases: verify the password (phase 1, no session) and establish the session (phase 2, after the second factor passes).

  ## Changes

  ### `app/libs/Auth/AuthService.php`
  - **`validateCredentials(string $username, string $password): User`** — validates credentials via `Auth::getProvider()->retrieveByCredentials()` without calling `Auth::attempt()`, so no session is established. Pre-checks the user's active state before touching the provider; if the account is locked it throws immediately with a distinct "is locked" message (the provider swallows `AuthenticationLockedUserLoginAttempt` and returns `null`, masking the lock state). Security checkpoints such as `LockUserCounterMeasure` still fire on provider-level failures.
  - **`loginUser(User $user, bool $remember): void`** — thin wrapper around `Auth::login()` for use after the second factor is verified.

  ### `app/libs/Utils/Services/IAuthService.php`
  - Added `validateCredentials()` and `loginUser()` to the interface with full doc-blocks explaining the session-safety contract.

  ### Tests
  - **`tests/unit/AuthServiceValidateCredentialsTest.php`** — isolated unit tests (Mockery alias mocks for `Auth` and `Log` facades, run in separate processes). Covers: valid credentials
  return user without session, invalid credentials throw `AuthenticationException`, locked account short-circuits before calling the provider, active user does not take the locked path, `loginUser()` with `remember = true/false`.
  - **`tests/AuthServiceValidateCredentialsIntegrationTest.php`** — integration test against the real database and security-checkpoint stack. Verifies that a failed `validateCredentials()` call increments `login_failed_attempt` via `LockUserCounterMeasure`, and that a successful call returns the `User` entity without establishing a session (`Auth::check()` stays `false`).

  ## Test plan

  - [x] Run unit suite: `docker compose exec app ./vendor/bin/phpunit tests/unit/AuthServiceValidateCredentialsTest.php`
  - [x] Run integration test: `docker compose exec app ./vendor/bin/phpunit tests/AuthServiceValidateCredentialsIntegrationTest.php`
  - [x] Run full suite to check for regressions: `docker compose exec app ./vendor/bin/phpunit`

--- 

### Current state

AuthService::login() validates credentials AND establishes a session in a single call. There is no way to validate a user's password without logging them in, which is required to insert a 2FA gate between credential validation and session establishment.

### Target state

AuthService exposes a new validateCredentials(string $username, string $password): User method that validates the password via the existing CustomAuthProvider::retrieveByCredentials() (which handles attempt tracking and account locking via security checkpoints) and returns the authenticated User object WITHOUT establishing a
session. Throws AuthenticationException on failure.

### TASKS

  - Add validateCredentials() method to IAuthService interface
  - Implement validateCredentials() in AuthService: delegates to Auth::getProvider()->retrieveByCredentials() with username and password, throws AuthenticationException if user is null
  - Add loginUser(User $user, bool $remember) method to IAuthService interface (needed by verify2FA to establish session after 2FA succeeds) → it should call  Auth::login($user, $remember);
  - Write unit test: valid credentials return User object without session being established
  - Write unit test: invalid credentials throw AuthenticationException
  - Write unit test: locked account throws AuthenticationException with locked-account message
  - Write integration test: validateCredentials() triggers security checkpoint tracking (failed attempt counter increments on failure)

 ### ACCEPTANCE CRITERIA

  - validateCredentials() with correct username/password returns the User object
  - validateCredentials() with wrong password throws AuthenticationException
  - validateCredentials() does NOT call Auth::login() or establish any session
  - After validateCredentials() succeeds, Session::get('login_web') is null (no authenticated session)
  - Security checkpoints still fire: failed attempts increment the counter, account locking still triggers after threshold
  - loginUser() establishes a Laravel session for the given User object ( call Auth::login() )
  - All tests pass

  ### DEVELOPMENT NOTES

  Key files:
  - Modified: app/Services/Auth/IAuthService.php (interface)
  - Modified: app/Services/Auth/AuthService.php (implementation)
  - Modified: app/libs/Auth/CustomAuthProvider.php (reference only, do not modify)
  - New: tests/Unit/AuthServiceValidateCredentialsTest.php

### Gotchas:
  - CustomAuthProvider::retrieveByCredentials() is the ONLY path for credential validation. Do NOT duplicate password checking logic. The provider handles password hashing, security checkpoints,
  and lock countermeasures.
  - The existing login() method calls Auth::attempt() which both validates AND establishes a session. validateCredentials() must NOT use Auth::attempt().
  - LockUserCounterMeasure tracks failed attempts. Ensure validateCredentials() triggers this on failure by going through the same code path as the existing login flow.

### Out of scope: 
Controller changes


app/libs/Auth/AuthService.php
```diff
@@ -149,11 +152,10 @@ public function login(string $username, string $password, bool $remember_me): bo
     {
         Log::debug("AuthService::login");
 
-        $this->last_login_error = "";
```
martinquiroga-exo: Why was this removed?

app/libs/Auth/AuthService.php
```diff
@@ -173,6 +175,54 @@ public function login(string $username, string $password, bool $remember_me): bo
...
+     * @return User
+     * @throws AuthenticationException
+     */
+    public function validateCredentials(string $username, string $password): User
+    {
+        Log::debug("AuthService::validateCredentials");
+
+        // retrieveByCredentials swallows AuthenticationLockedUserLoginAttempt and returns null,
+        // so pre-check lock state here to surface a distinct message for locked accounts.
+        $existing = $this->user_repository->getByEmailOrName($username);
```
martinquiroga-exo: As stated a few lines below `retrieveByCredentials()` already calls `getByEmailOrName()` which is an unnecessary double db call.
Moreover inside `retrieveByCredentials()` we are already checking this: https://github.com/OpenStackweb/openstackid/blob/db9f77784c442f785b57c44ac6a846d015bfe6a6/app/libs/Auth/CustomAuthProvider.php#L145

I believe the real issue lays in that all Exceptions are been swallowed by this: https://github.com/OpenStackweb/openstackid/blob/db9f77784c442f785b57c44ac6a846d015bfe6a6/app/libs/Auth/CustomAuthProvider.php#L164

Since we are already throwing in most places I would be inclined to remove this and have a proper flow. @smarcet do you agree?

app/libs/Auth/AuthService.php
```diff
@@ -173,6 +175,54 @@ public function login(string $username, string $password, bool $remember_me): bo
...
+
+    /**
+     * @param User $user
+     * @param bool $remember
+     * @return void
+     */
+    public function loginUser(User $user, bool $remember): void
+    {
+        Log::debug("AuthService::loginUser");
+        Auth::login($user, $remember);
```
martinquiroga-exo, line 455: To avoid a race condition please do a check before calling `Auth:login(...)`

martinquiroga-exo (changes requested):
@matiasperrone-exo please see comments

app/libs/Auth/AuthService.php
smarcet: nope i dont agree is out of the scope to change existing code.
that will break security checkpoints mechanism already in place please re read the GOAL of the ticket
````
AuthService exposes a new validateCredentials(string $username, string $password): User method that validates the password via the existing CustomAuthProvider::retrieveByCredentials() (which handles attempt tracking and account locking via security checkpoints) and returns the authenticated User object WITHOUT establishing a
session. Throws AuthenticationException on failure.
````
validateCredentials() does not match the ticket contract 

1.Remove the pre-check entirely.
2. Call Auth::getProvider()->retrieveByCredentials([...]), then if (is_null($user) || !$user instanceof User || !$user->canLogin()) throw new        
  AuthenticationException(...).
3.  If the locked-state UX matters, expose it through the provider/exception path, not via a duplicated lookup.

File: app/libs/Auth/AuthService.php [hunk 4]
```diff
@@ -149,20 +152,19 @@ public function login(string $username, string $password, bool $remember_me): bo
     {
         Log::debug("AuthService::login");
 
-        $this->last_login_error = "";
         if (!Auth::attempt(['username' => $username, 'password' => $password], $remember_me)) {
             throw new AuthenticationException
             (
-                "We are sorry, your username or password does not match an existing record."
+                "username or password does not match an existing record."
```
smarcet: @martinquiroga-exo @matiasperrone-exo changes to the existent code is out of scope

app/libs/Auth/AuthService.php
```diff
@@ -149,20 +152,19 @@ public function login(string $username, string $password, bool $remember_me): bo
...
         }
         Log::debug("AuthService::login: clearing principal");
         $this->principal_service->clear();
         $current_user = $this->getCurrentUser();
-        if(is_null($current_user) || !$current_user->canLogin())
+        if (is_null($current_user) || !$current_user->canLogin())
             throw new AuthenticationException
             (
-                "We are sorry, your username or password does not match an existing record."
+                "username or password does not match an existing record."
```
smarcet: @martinquiroga-exo @matiasperrone-exo changes to existing code is out of scope

app/libs/Auth/AuthService.php
smarcet: re-check the user immediately before Auth::login(): if (!$user->isActive() || !$user->canLogin()) throw new AuthenticationException(...

app/libs/Auth/AuthService.php
```diff
@@ -135,7 +134,11 @@ public function isUserLogged()
      */
     public function getCurrentUser(): ?User
     {
-        return Auth::user();
+        $user = Auth::user();
```
smarcet: @matiasperrone-exo revert this
changes to the original code are out of scope
silently converts non-User authenticated principals to null
Pre-PR returned whatever Auth::user() returned; PR now returns null if the principal isn't instanceof User. ~25 callers (ClientService, TokenService, strategies, middleware, controllers) immediately deref the result and would break loudly before, but now silently treat the request as anonymous. This is a behavior change in an existing method, again outside the ticket's stated scope

smarcet (changes requested):
@matiasperrone-exo please review comments
@matiasperrone-exo @martinquiroga-exo changes on existent code are out of scope
anything like that should be discussed before any proposed changes

app/libs/Auth/AuthService.php
matiasperrone-exo: Because it was dead code.
Restored

app/libs/Auth/AuthService.php
matiasperrone-exo: Ok moved to:

```php
   /**
     * @param string $username
     * @param string $password
     * @return User
     * @throws AuthenticationException
     */
    public function validateCredentials(string $username, string $password): User
    {
        Log::debug("AuthService::validateCredentials");

        /**
         * @var User|null $user
         */
        $user = $this->user_repository->getByEmailOrName($username);
        $valid = Auth::getProvider()->validateCredentials($user, ['username' => $username, 'password' => $password]);
        if (!$valid) {
            throw new AuthenticationException();
        }

        return $user;
    }
```

app/libs/Auth/AuthService.php
matiasperrone-exo: reverted

app/libs/Auth/AuthService.php
matiasperrone-exo: reverted

app/libs/Auth/AuthService.php
matiasperrone-exo: Reverted

app/libs/Auth/AuthService.php
matiasperrone-exo: done

martinquiroga-exo (approved):
LGTM

smarcet:
@martinquiroga-exo the ticket ask for 
````
Implement validateCredentials() in AuthService: delegates to Auth::getProvider()->retrieveByCredentials() with username and password, throws AuthenticationException if user is null
````

but the code is not following that please re review it 

https://github.com/OpenStackweb/openstackid/pull/127/changes#diff-31b8c36cbff745e9a0410953fc435b4b933632a787b6815b2e861af3a5f6adb3R198

**Review by smarcet — CHANGES_REQUESTED** [review]
@martinquiroga-exo @matiasperrone-exo please re review

File: app/libs/Auth/AuthService.php (hunk 2 above)
Comment (smarcet) [reply app/libs/Auth/AuthService.php]: @matiasperrone-exo this is not what the ticket asked for please re read it
u are using now Auth::getProvider()->validateCredentials and the ticket clearly states 
```
Implement validateCredentials() in AuthService: delegates to Auth::getProvider()->retrieveByCredentials() with username and password, throws AuthenticationException if user is null
```

File: app/libs/Auth/AuthService.php [hunk 7]
```diff
@@ -173,6 +173,39 @@ public function login(string $username, string $password, bool $remember_me): bo
...
+     */
+    public function validateCredentials(string $username, string $password): User
+    {
+        Log::debug("AuthService::validateCredentials");
+
+        /**
+         * @var User|null $user
+         */
+        $user = Auth::getProvider()->retrieveByCredentials(['username' => $username, 'password' => $password]);
+        if (!$user) {
```
caseylocker: The guard here doesn't match what was requested in the prior review — the prescription was:

```php
if (is_null($user) || !$user instanceof User || !$user->canLogin())
    throw new AuthenticationException(...);
```

HEAD only checks `if (!$user)`, so a non-`User` principal returned by a custom provider, or a `User` whose `canLogin()` is false, would slip through and be returned to the caller. `loginUser()` would then fail later than it should and with less information. Please add the `instanceof User` and `!$user->canLogin()` checks as requested.

app/libs/Auth/AuthService.php
```diff
@@ -173,6 +173,39 @@ public function login(string $username, string $password, bool $remember_me): bo
...
+     * @throws AuthenticationException
+     */
+    public function validateCredentials(string $username, string $password): User
+    {
+        Log::debug("AuthService::validateCredentials");
+
+        /**
+         * @var User|null $user
+         */
+        $user = Auth::getProvider()->retrieveByCredentials(['username' => $username, 'password' => $password]);
```
caseylocker, line 431: `CustomAuthProvider::retrieveByCredentials()` rethrows `UnverifiedEmailMemberException` (`app/libs/Auth/CustomAuthProvider.php:159-162`) — it is *not* swallowed by the catch-all at `:164-168`. `UnverifiedEmailMemberException extends Exception` (not `AuthenticationException`), so when an unverified user attempts login the exception escapes `validateCredentials()` uncaught. That violates this method's contract — `IAuthService::validateCredentials()` declares `@throws AuthenticationException` only (`app/libs/Utils/Services/IAuthService.php:67-68`), and callers (the upcoming MFA controller in particular) won't have a `catch` for `UnverifiedEmailMemberException`.

Suggested fix, scoped to this new method only:

```php
try {
    $user = Auth::getProvider()->retrieveByCredentials(['username' => $username, 'password' => $password]);
} catch (UnverifiedEmailMemberException $ex) {
    throw new AuthenticationException($ex->getMessage());
}
if (is_null($user) || !$user instanceof User || !$user->canLogin()) {
    throw new AuthenticationException("We are sorry, your username or password does not match an existing record.");
}
return $user;
```

No changes to existing code — only the new method.

caseylocker (changes requested):
Two outstanding items before this can merge — both scoped to the new `validateCredentials()` method, no existing-code changes required:

1. The post-`retrieveByCredentials()` guard doesn't match what was requested in the prior review.
2. `UnverifiedEmailMemberException` escapes `validateCredentials()` uncaught, violating the `IAuthService` contract.

Inline comments below. Also worth a quick rewrite of the PR description before merge — it still describes the pre-check and `getCurrentUser()` `instanceof User` guard that were reverted in 70ba198e, so the merged record won't match the diff.

app/libs/Auth/AuthService.php
matiasperrone-exo: done

app/libs/Auth/AuthService.php
matiasperrone-exo: done

caseylocker (approved):
LGTM

app/libs/Auth/AuthService.php
matiasperrone-exo: done

PR #128, opened 2026-05-07
fix(oauth2): allow HTTP origins and fix trailing-slash mismatch for l…

mulldug:
…ocal dev

- Strip trailing slash from normalizeUrl() results in isOriginAllowed() so origins with/without trailing slash match registered values
- Conditionally skip HTTPS scheme enforcement when SSL.Enable is false, allowing HTTP clients in local dev
- Fix SSL.Enable config key lookup (server.SSL_Enable → server.ssl_enable) to match the key name defined in config/server.php
- Use getSchemeAndHttpHost() instead of hardcoded https:// prefix so the IDP base URL reflects the actual request scheme
- Add fastcgi_buffers directives to nginx config to prevent buffer overflow

### DIFF
--- a/app/Models/OAuth2/Client.php
+++ b/app/Models/OAuth2/Client.php
@@ -809,9 +809,9 @@ class Client extends BaseEntity implements IClient
     {
         $originWithoutPort = URLUtils::canonicalUrl($origin, false);
         if(empty($originWithoutPort)) return false;
-        if(str_contains($this->allowed_origins, URLUtils::normalizeUrl($originWithoutPort) )) return true;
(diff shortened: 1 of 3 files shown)


config/server.php
```diff
@@ -15,6 +15,7 @@
 return array
 (
     'ssl_enabled'     => env('SSL_ENABLED', false),
+    'ssl_enable'      => env('SSL_ENABLED', false),
```
smarcet: this is duping "server.ssl_enabled"
please use old key

mulldug: Done.

PR #133, opened 2026-05-21
Feature | Add Device Trust Service

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/86ba0nah4

## Blocked by:

PR: [Feature | MFA Challenge Strategy Pattern (Interface, Abstract, Factory, EmailOTP) #129](#129)

### New service layer (`app/Services/Auth/`)

  - **`IDeviceTrustService.php`** — Interface defining the contract: `trustDevice`, `isDeviceTrusted`, `removeTrustedDevices`, `generateDeviceIdentifier`.
  - **`DeviceTrustService.php`** — Implementation:
    - `trustDevice`: generates a 128-char random hex token, hashes it with SHA-256, persists a `UserTrustedDevice` record with configurable expiry (default 30 days), returns the raw token
  for cookie storage.
    - `isDeviceTrusted`: hashes the cookie token, looks up the record, returns `false` if not found / revoked / expired; updates `last_seen_at` on a valid hit.
    - `removeTrustedDevices`: bulk-revokes all devices for a user (sets `is_revoked = true`).
  - **`TwoFactorServiceProvider.php`** — Deferred service provider that binds `IDeviceTrustService → DeviceTrustService` as a singleton; registered in `config/app.php`.

### Repository changes (`IUserTrustedDeviceRepository` / `DoctrineUserTrustedDeviceRepository`)

  Two new methods added:

  - `getByUserAndDeviceIdentifier` — looks up a device by user + hashed identifier with no revoked/expiry filter (used by the service to check all states).
  - `revokeAllForUser` — bulk DQL `UPDATE` setting `is_revoked = true` for all devices belonging to a user.

### Entity updates (`UserTrustedDevice`)

  - Constructor now initializes `last_seen_at` to `now()`.
  - Added `isExpired()` helper (compares `expires_at` against now).
  - Fixed `targetEntity` reference from FQCN string to `User::class`.

### Config

  - **`config/two_factor.php`** — Added `device_trust_lifetime_days` (default `30`) and `cookie_name` (default `'device_trust_token'`) settings, driven by env vars.

### Tests (`tests/DeviceTrustServiceTest.php`)

  264-line unit test suite using Mockery covering:

  - `isDeviceTrusted`: null/empty cookie, unknown token, revoked device, expired device, valid device, `last_seen_at` update.
  - `trustDevice`: returns 128-char hex token, stores SHA-256 hash (not raw token), persists exactly one record.
  - `removeTrustedDevices`: delegates to `revokeAllForUser`.
  - `generateDeviceIdentifier`: SHA-256 correctness.

----

### Current state

No device trust mechanism exists. Users would need to complete 2FA on every single login, even from the same browser and device they used yesterday.

### Target state

`IDeviceTrustService` and its implementation allow the system to remember trusted devices via a secure cookie mechanism. A SHA-256 hash of a cryptographically random token is stored
in the user_trusted_devices table with a configurable TTL (default 30 days). The raw token is stored in a long-lived HttpOnly cookie , On subsequent logins, the cookie token is hashed and compared against stored records to bypass 2FA.

### TASKS

  - Create `IDeviceTrustService` interface with methods: `isDeviceTrusted(User, ?string cookieToken): bool`, `trustDevice(User, string userAgent, string ipAddress): string` (returns raw cookie token), `removeTrustedDevices(User): void`, `generateDeviceIdentifier(string token): string` (returns SHA-256 hash)
  - Implement `DeviceTrustService`: `isDeviceTrusted()` hashes the cookie token via `generateDeviceIdentifier()`, queries `IUserTrustedDeviceRepository` for a matching non-revoked, non-expired record for the user, updates `last_seen_at` on match.
  - `trustDevice()` generates a 64-byte random token via `bin2hex(random_bytes(32))`, stores SHA-256 hash in user_trusted_devices with TTL from config, extracts User-Agent for `device_name`, returns raw token.
  - `removeTrustedDevices()` revokes all devices for the user (sets `is_revoked=true`).
  - Implement `IUserTrustedDeviceRepository::getByUserAndDeviceIdentifier()` query method
  - Register `IDeviceTrustService` in `TwoFactorServiceProvider`
  - Write unit tests for `isDeviceTrusted()` with: valid trusted device, expired device, revoked device, no cookie, wrong cookie
  - Write unit tests for `trustDevice()`: verify token generation, hash storage, and record creation
  - Write unit tests for `removeTrustedDevices()`: verify all devices for user are revoked
- `DeviceTrustService` must depend on `IUserTrustedDeviceRepository`, not `EntityManager` directly.
- `trustDevice()` must instantiate a `UserTrustedDevice` entity, associate it with the `User`, set `device_identifier`, `device_name`, `ip_address`, `user_agent`, `trusted_at`, `expires_at`, `last_seen_at`, `is_revoked=false`, then persist it through the repository.
- `isDeviceTrusted()` must call `IUserTrustedDeviceRepository::getByUserAndDeviceIdentifier($user, $deviceIdentifier)` and only return true for a non-revoked, non-expired record.
- On successful match, `isDeviceTrusted()` must update `last_seen_at` and persist the change.
- `removeTrustedDevices()` must call a repository method that revokes all trusted devices for the user.

### ACCEPTANCE CRITERIA

- `trustDevice()` returns a 128-character hex string (64 bytes as hex)
- The stored `device_identifier` is a SHA-256 hash of the returned token (not the raw token itself)
- `isDeviceTrusted()` returns true when a matching non-revoked, non-expired record exists
- `isDeviceTrusted()` returns `false` for expired records (`expires_at < now`)
- `isDeviceTrusted()` returns `false` for revoked records (`is_revoked = true`)
- `isDeviceTrusted()` returns `false` when cookieToken is null or empty
- `isDeviceTrusted()` updates `last_seen_at` on a valid match
- `removeTrustedDevices()` sets is_revoked=true on all records for the user
- DeviceTrustService does not query Doctrine directly; all `UserTrustedDevice` access goes through IUserTrustedDeviceRepository.
- `trustDevice()` creates exactly one `UserTrustedDevice` record per call.
- The raw cookie token is returned to the caller but is never stored in `UserTrustedDevice`.
- `UserTrustedDevice::device_identifier` stores hash('sha256', $rawToken).
- `isDeviceTrusted()` updates `UserTrustedDevice::last_seen_at` when a valid device is found.
- Device trust TTL defaults to 30 days, configurable via `config('two_factor.device_trust_lifetime_days')`
- cookie name is configuravle via a `config('two_factor.cookie_name')`
- All unit tests pass

### DIFF
--- a/app/Repositories/DoctrineUserTrustedDeviceRepository.php
+++ b/app/Repositories/DoctrineUserTrustedDeviceRepository.php
@@ -31,6 +31,30 @@ final class DoctrineUserTrustedDeviceRepository
         return Criteria::expr()->gt('expires_at', $now);
     }
 
+    public function getByUserAndDeviceIdentifier(User $user, string $deviceIdentifier): ?UserTrustedDevice
+    {
+        $criteria = Criteria::create()
+            ->where(Criteria::expr()->eq('user', $user))
+            ->andWhere(Criteria::expr()->eq('device_identifier', $deviceIdentifier))
+            ->setMaxResults(1);
+
+        $result = $this->matching($criteria)->first();
+        return $result instanceof UserTrustedDevice ? $result : null;
+    }
+
+    public function revokeAllForUser(User $user): void
+    {
+        $this->getEntityManager()
+            ->createQueryBuilder()
+            ->update($this->getBaseEntity(), 'd')
+            ->set('d.is_revoked', ':revoked')
+            ->where('d.user = :user')
+            ->setParameter('revoked', true)
+            ->setParameter('user', $user)
+            ->getQuery()
+            ->execute();
+    }
+
     public function getActiveByUserAndIdentifier(User $user, string $deviceIdentifier): ?UserTrustedDevice
     {
         $criteria = Criteria::create()
--- a/app/Services/Auth/DeviceTrustService.php
+++ b/app/Services/Auth/DeviceTrustService.php
@@ -1,0 +1,85 @@
+<?php
+namespace App\Services\Auth;
+/**
+ * Copyright 2025 OpenStack Foundation
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+use App\libs\Auth\Models\UserTrustedDevice;
+use Auth\Repositories\IUserTrustedDeviceRepository;
+use Auth\User;
+use DateTime;
+use DateInterval;
+use DateTimeZone;
+
+/**
+ * Class DeviceTrustService
+ * @package App\Services\Auth
+ */
+final class DeviceTrustService implements IDeviceTrustService
+{
+    public function __construct(private readonly IUserTrustedDeviceRepository $repository)
+    {
+    }
+
+    public function generateDeviceIdentifier(string $token): string
+    {
+        return hash('sha256', $token);
+    }
+
+    public function trustDevice(User $user, string $userAgent, string $ipAddress): string
+    {
+        $rawToken = bin2hex(random_bytes(64));
+
+        $lifetimeDays = (int) config('two_factor.device_trust_lifetime_days', 30);
+        $now = new DateTime('now', new DateTimeZone('UTC'));
+        $expiresAt = clone $now;
+        $expiresAt->add(new DateInterval("P{$lifetimeDays}D"));
+
+        $device = new UserTrustedDevice();
+        $device->setUser($user);
+        $device->setDeviceIdentifier($this->generateDeviceIdentifier($rawToken));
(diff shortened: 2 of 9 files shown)


caseylocker (approved):
LGTM. All Clickup tasks/requirements are met. One suggestion which is @smarcet call:

`tests/DeviceTrustServiceTest.php:29` extends `BrowserKitTestCase`, whose `setUp()` runs a full Redis flush, doctrine migrations, and `TestSeeder` on every test method
  (`tests/BrowserKitTestCase.php:36-56`). This suite is pure Mockery and needs none of that — it pays heavy CI cost per method and risks the pattern getting copied as more 2FA service tests land.

  Suggest extending `Tests\TestCase` (`tests/TestCase.php:19`) instead. That still boots Laravel so `config('two_factor.device_trust_lifetime_days')` resolves in
  `testTrustDeviceSetsExpiresAtFromConfig`, but skips the migrate + seed.

  While you're in there, three small tightening opportunities on the same file:

  - `:72` — `shouldReceive('getByUserAndDeviceIdentifier')` has no `with(...)`; add `->with($user, hash('sha256', 'unknowntoken'))` so the test proves the lookup uses the SHA-256 of the cookie, not the raw token.
  - `:118` — `shouldReceive('add')->once()` doesn't verify the `$sync=true` second arg. Add `->with(Mockery::type(UserTrustedDevice::class), true)` so a regression that drops the sync flag is caught.
  - `:230` — `$lifetimeDays = (int) config(...)` makes the assertion self-referential (passes for any config value, including 0). Use  Config::set('two_factor.device_trust_lifetime_days', 45)` at the top of the test and assert `45` literally.

PR #136, opened 2026-06-03
Feature | UserController MFA Integration, Device Trust Cookie Management, Audit Wiring, and 2FA Rate Limiting

matiasperrone-exo:
## Task:

Ref: https://app.clickup.com/t/9014802374/86ba2zc6p

  ### What Changed

  #### `UserController` (`app/Http/Controllers/UserController.php`)
  - Injected `IDeviceTrustService`, `ITwoFactorAuditService`, and `ITwoFactorGateService` via constructor.
  - **Password flow** now validates credentials without creating a session first. If `MFAGateService::requiresChallenge()` returns `true`, issues a challenge (via
  `AuthService::issueMFAChallenge`) and returns `{ error_code: "mfa_required", ... }` — no session created yet.
  - Added `resolveClientFromMemento()` private helper to extract the OAuth2 client from a pending session memento (deduplicates logic previously repeated in password and OTP flows).
  - Added `postVerify2FA()`, `postResend2FA()`, and `postRecovery2FA()` action methods to handle the full MFA step-up login lifecycle.
  - Applied `MFACookieManager` trait for reading/writing the trusted-device cookie.

  #### `MFACookieManager` trait (`app/Http/Controllers/Traits/MFACookieManager.php`) — **new file**
  - Provides `getCookieToken(): ?string` (reads raw token from cookie).
  - Provides `queueDeviceTrustCookie(User $user): void` (calls `IDeviceTrustService::trustDevice` and queues a secure, HttpOnly cookie).
  - Contains **no Doctrine/repository logic** — cookie transport only.

  #### `TwoFactorRateLimitMiddleware` (`app/Http/Middleware/TwoFactorRateLimitMiddleware.php`) — **new file**
  - Throttles `verify`, `recovery`, and `resend` MFA endpoints.
  - Counters stored in **cache** (not session) with a fixed-window TTL.
  - `verify` / `recovery`: increments only on failed response (`mfa_verification_failed`, `mfa_invalid_recovery`).
  - `resend`: increments on every request.
  - Returns `HTTP 429` with `error_code: mfa_rate_limit` when the limit is exceeded.
  - Registered in `app/Http/Kernel.php` and wired to the new routes in `routes/web.php`.

  #### `AuthService` (`app/libs/Auth/AuthService.php`) and `IAuthService` (`app/libs/Utils/Services/IAuthService.php`)
  - Added `validateCredentials(string $username, string $password): User` — validates without creating a session.
  - Added `loginUser(User $user, bool $remember): void` — establishes the session for an already-validated user.
  - Added `issueMFAChallenge(User $user, MFAChallengeStrategy $strategy, ?Client $client, bool $remember): array` — triggers the challenge strategy and stores pending state.

  #### `DeviceTrustService` and `TwoFactorAuditService` (minor fixes)
  - Resolved wiring/dependency issues surfaced during integration testing.

  #### `config/two_factor.php` — **new file**
  - Centralizes MFA configuration: `cookie_name`, `device_trust_lifetime_days`, rate-limit thresholds per action.

  #### `EncryptCookies` middleware
  - Excluded the `device_trust_token` cookie from encryption (raw HMAC token must be read as-is by `IDeviceTrustService`).

  ---

  ### Tests

  | File | What it covers |
  |---|---|
  | `tests/TwoFactorLoginFlowTest.php` (**new**) | End-to-end: password→MFA challenge→verify→session; device trust skip; rate limiting; recovery codes; resend. |
  | `tests/DeviceTrustServiceTest.php` | Expanded with cookie integration assertions. |
  | `tests/Unit/TwoFactorAuditServiceTest.php` | Expanded with new event types. |
  | `tests/Unit/MFA/EmailOTPMFAChallengeStrategyTest.php` | Added cases for `issueMFAChallenge` path. |
  | `phpunit.xml` | Registered **Two Factor Authentication Test Suite**. |

### DIFF
--- a/app/Http/Controllers/UserController.php
+++ b/app/Http/Controllers/UserController.php
@@ -17,7 +17,17 @@ use RyanChandler\LaravelCloudflareTurnstile\Rules\Turnstile;
 use App\Jobs\RevokeUserGrantsOnExplicitLogout;
 use App\Http\Controllers\OpenId\OpenIdController;
 use App\Http\Controllers\Traits\JsonResponses;
+use App\Http\Controllers\Traits\MFACookieManager;
 use App\Http\Utils\CountryList;
+use App\libs\Auth\Models\TwoFactorAuditLog;
+use App\Services\Auth\IDeviceTrustService;
+use App\Services\Auth\ITwoFactorAuditService;
+use App\Services\Auth\ITwoFactorGateService;
+use Auth\User;
+use Models\OAuth2\Client;
+use Strategies\MFA\MFAChallengeStrategyFactory;
+use Symfony\Component\HttpFoundation\Response as HttpResponse;
+use Utils\IPHelper;
 use App\libs\OAuth2\Strategies\LoginHintProcessStrategy;
 use App\ModelSerializers\SerializerRegistry;
 use Auth\Exceptions\AuthenticationException;
@@ -133,6 +143,21 @@ final class UserController extends OpenIdController
     private $security_context_service;
 
     /**
+     * @var IDeviceTrustService
+     */
+    private $device_trust_service;
+
+    /**
+     * @var ITwoFactorAuditService
+     */
+    private $two_factor_audit_service;
+
+    /**
+     * @var ITwoFactorGateService
+     */
+    private $mfa_gate_service;
+
+    /**
      * @param IMementoOpenIdSerializerService $openid_memento_service
      * @param IMementoOAuth2SerializerService $oauth2_memento_service
      * @param IAuthService $auth_service
@@ -167,7 +192,10 @@ final class UserController extends OpenIdController
         IResourceServerService $resource_server_service,
         IUtilsServerConfigurationService $utils_configuration_service,
         ISecurityContextService $security_context_service,
-        LoginHintProcessStrategy $login_hint_process_strategy
+        LoginHintProcessStrategy $login_hint_process_strategy,
+        IDeviceTrustService $device_trust_service,
+        ITwoFactorAuditService $two_factor_audit_service,
+        ITwoFactorGateService $mfa_gate_service,
     )
     {
         $this->openid_memento_service = $openid_memento_service;
@@ -185,6 +213,9 @@ final class UserController extends OpenIdController
         $this->resource_server_service = $resource_server_service;
         $this->utils_configuration_service = $utils_configuration_service;
         $this->security_context_service = $security_context_service;
+        $this->device_trust_service = $device_trust_service;
+        $this->two_factor_audit_service = $two_factor_audit_service;
+        $this->mfa_gate_service = $mfa_gate_service;
 
         $this->middleware(function ($request, $next) use($login_hint_process_strategy){
 
@@ -253,6 +284,8 @@ final class UserController extends OpenIdController
     }
 
     use JsonResponses;
+
+    use MFACookieManager;
 
     /**
      * @return \Illuminate\Http\JsonResponse|mixed
@@ -436,35 +469,41 @@ final class UserController extends OpenIdController
                 $connection = $data['connection'] ?? null;
 
                 try {
-                    if ($flow == "password" && $this->auth_service->login($username, $password, $remember)) {
+                    if ($flow == "password") {
+                        // Validate credentials WITHOUT establishing a session, so the
+                        // MFA gate can run before the user is authenticated.
+                        $user = $this->auth_service->validateCredentials($username, $password);
+
+                        $cookieToken = $this->getCookieToken();
+
+                        if ($this->mfa_gate_service->requiresChallenge($user, $cookieToken)) {
+                            // Issue a challenge and stop short of session creation.
+                            $client   = $this->resolveClientFromMemento();
+                            $method   = $user->getTwoFactorMethod();
+                            $strategy = MFAChallengeStrategyFactory::create($method);
+                            $payload = $this->auth_service->issueMFAChallenge($user, $strategy, $client, $remember);
+
+                            $this->two_factor_audit_service->log(
+                                $user,
+                                TwoFactorAuditLog::EventChallengeIssued,
+                                $method,
+                                IPHelper::getUserIp()
+                            );
+
+                            return Response::json(
+                                array_merge(['error_code' => 'mfa_required'], $payload),
+                                HttpResponse::HTTP_OK
+                            );
+                        }
+
+                        // No challenge required: establish the session and continue.
+                        $this->auth_service->loginUser($user, $remember);
                         return $this->login_strategy->postLogin();
                     }
 
                     if ($flow == "otp") {
 
-                        $client = null;
-
-                        // check if we have a former oauth2 request
-                        if ($this->oauth2_memento_service->exists()) {
-
-                            Log::debug("UserController::postLogin exist a oauth auth request on session");
-
-                            $oauth_auth_request = OAuth2AuthorizationRequestFactory::getInstance()->build
-                            (
-                                OAuth2Message::buildFromMemento($this->oauth2_memento_service->load())
-                            );
-
-                            if ($oauth_auth_request->isValid()) {
-
-                                $client_id = $oauth_auth_request->getClientId();
-
-                                $client = $this->client_repository->getClientById($client_id);
-                                if (is_null($client))
-                                    throw new ValidationException("client does not exists");
-
-                                $this->oauth2_memento_service->serialize($oauth_auth_request->getMessage()->createMemento());
-                            }
-                        }
+                        $client = $this->resolveClientFromMemento();
 
                         $otpClaim = OAuth2OTP::fromParams($username, $connection, $password);
                         $this->auth_service->loginWithOTP($otpClaim, $client);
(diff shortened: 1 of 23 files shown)


app/Http/Controllers/UserController.php
```diff
@@ -435,35 +468,41 @@ public function postLogin()
...
-
-                        // check if we have a former oauth2 request
-                        if ($this->oauth2_memento_service->exists()) {
-
-                            Log::debug("UserController::postLogin exist a oauth auth request on session");
-
-                            $oauth_auth_request = OAuth2AuthorizationRequestFactory::getInstance()->build
-                            (
-                                OAuth2Message::buildFromMemento($this->oauth2_memento_service->load())
+                    if ($flow == "password") {
```
smarcet: @matiasperrone-exo please move flow to constants on 
IAuthService::AuthenticationFlowPassword
and 
IAuthService::AuthenticationFlowOTP

app/Http/Controllers/UserController.php
```diff
@@ -435,35 +468,41 @@ public function postLogin()
...
-                            if ($oauth_auth_request->isValid()) {
+                            return Response::json(
+                                array_merge(['error_code' => 'mfa_required'], $payload),
+                                HttpResponse::HTTP_OK
+                            );
+                        }
 
-                                $client_id = $oauth_auth_request->getClientId();
+                        // No challenge required: establish the session and continue.
+                        $this->auth_service->loginUser($user, $remember);
```
caseylocker, line 587: **CRITICAL: login no longer registers the OIDC principal.**

This replaces the previous `auth_service->login()` call, but `loginUser()` (`app/libs/Auth/AuthService.php:445`) only calls `Auth::login()` — it does not call `principal_service->clear()` + `register($userId, time())` the way `login()` does (`AuthService.php:402-414`). Nothing else re-registers it (the `Login` event listener `OnUserLogin` only dispatches a deferred `PostLoginUser` job).

Consequences for the interactive SSO flow: `PrincipalService::get()->getAuthTime()` is null, so the id_token `auth_time` is wrong/0 (`TokenService.php:335`), `InteractiveGrantType::shouldForceReLogin` (`:668`) gates re-auth on `getAuthTime() > 0` and therefore silently skips `max_age` / `prompt=login` enforcement, and OP browser state / session management breaks. This affects all three new `loginUser()` call sites — here (non-MFA), `:686` (verify2FA), and `:761` (recovery) — i.e. every password login, not just MFA.

Fix: have `loginUser()` mirror `login()`'s post-auth principal registration (clear + register with auth time and OP browser state).

app/Strategies/MFA/EmailOTPMFAChallengeStrategy.php
```diff
@@ -54,10 +60,12 @@ public function verifyChallenge(User $user, string $code, ?Client $client = null
         }
 
         $otp->redeem();
+        $this->otp_repository->add($otp, false);
```
caseylocker: **OTP redeem lacks the pessimistic row lock (TOCTOU double-redeem).**

`verifyChallenge` fetches via `getByValueConnectionAndUserName` (plain SELECT) then `redeem()` at `:62`, without acquiring `IOAuth2OTPRepository::refreshExclusiveLock()` (`DoctrineOAuth2OTPRepository.php:142`). With the transaction service defaulting to READ COMMITTED (`DoctrineTransactionService.php:133`), two concurrent submissions of the same valid code can both read the unredeemed row and both succeed.

SDS §4.5 and the ticket gotcha explicitly say to prefer delegating to `AuthService::verifyOTPChallenge()` (tx-managed, value-checked, row-locked, session-bound) rather than reimplementing redemption. Either route through `verifyOTPChallenge()` or call `refreshExclusiveLock()` and re-check redemption state before `redeem()`.

app/libs/Auth/AuthService.php
```diff
@@ -778,4 +774,46 @@ public function postLoginUserActions(int $user_id): void
...
+        });
+    }
+
+    public function verifyMFAChallenge(
+        User $user,
+        IMFAChallengeStrategy $strategy,
+        string $value
+    ): void {
+        $this->tx_service->transaction(function () use ($user, $strategy, $value) {
+            $strategy->verifyChallenge($user, $value);
```
caseylocker: **MFA OTP is issued with a client but verified with `client = null`.**

The challenge is issued with the resolved client (`UserController.php:480`), but `verifyMFAChallenge` calls `verifyChallenge($user, $value)` with no client, so the lookup skips client scoping (`DoctrineOAuth2OTPRepository.php:130`). More concretely, the sibling-revocation in the strategy (`EmailOTPMFAChallengeStrategy.php:65`) calls `getByUserNameNotRedeemed($email)` without a client, so verifying an MFA OTP redeems all of the user's pending email OTPs — including unrelated passwordless-login OTPs for other clients.

Fix: carry the client id in the pending MFA state and pass the same client into verification (and into the sibling-revoke query), or delegate to `verifyOTPChallenge($claim, $user, $client)`.

app/Strategies/MFA/AbstractMFAChallengeStrategy.php
```diff
@@ -51,6 +51,7 @@ public function verifyRecoveryCode(User $user, string $code): void
         foreach ($this->recovery_code_repository->getUnusedByUser($user) as $recoveryCode) {
             if (Hash::check($code, $recoveryCode->getCodeHash())) {
                 $recoveryCode->markUsed();
+                $this->recovery_code_repository->add($recoveryCode, false);
```
caseylocker: **Recovery-code redemption has a double-spend race.**

`verifyRecoveryCode` iterates `getUnusedByUser` (`:51`), `Hash::check`s, then `markUsed()` + `add(false)` with no pessimistic lock or conditional update, under READ COMMITTED (`DoctrineTransactionService.php:133`). Two concurrent requests can both observe the same code unused and both succeed.

Fix: lock the user's unused recovery rows before hash-checking, or redeem via an atomic `UPDATE ... WHERE id = ? AND used_at IS NULL` after identifying the match.

app/Http/Controllers/UserController.php
```diff
@@ -557,6 +596,246 @@ public function postLogin()
...
+            // Second factor verified: establish the session.
+            $this->auth_service->loginUser($user, (bool) $pending['remember']);
+
+            if ($trust_device) {
+                $this->queueDeviceTrustCookie($user);
+            }
+
+            $strategy->clearPendingState();
+
+            $this->two_factor_audit_service->log(
```
caseylocker: **Post-verification side effects can 500 after the OTP is already burned.**

By this point the OTP redeem has committed (`verifyMFAChallenge`, `:670`) and the session is established (`loginUser`, `:686`). `queueDeviceTrustCookie` (`:689`, its own tx in `DeviceTrustService`) and this audit `log()` (its own tx in `TwoFactorAuditService`) run afterward; if either throws, the outer `catch (Exception)` returns `error500` even though the user is effectively logged in and the single-use OTP is spent — on retry they are locked out. If `trustDevice` throws, pending state is not cleared either.

Fix: make audit best-effort/non-blocking, and/or order durable side effects before `loginUser()`, keeping device-trust persistence consistent with the redeem boundary. Ticket Task 5 asks for exactly this decision ("redeem and device-trust persistence are atomic; audit may be best-effort").

caseylocker (changes requested):
Reviewed against ClickUp 86ba2zc6p and SDS `idp-mfa.md` (§4.4 controller flow, §4.5 strategies, §4.9.1 AuthService MFA methods).

The transaction-context design rule is correctly implemented — the controller orchestrates only, all strategy DB-mutating calls go through `AuthService::*MFA*` wrappers, and no `add(entity, true)` self-flush remains. Every acceptance criterion is met on the happy path, and `verifyChallenge` now correctly loads the persisted OTP and compares its value.

However, five issues should be addressed before merge. All five are consequences of work first made live in this PR (the MFA strategies and `loginUser()` were previously unwired/dead code), so they are in scope here rather than the base PR. The first is a critical regression in the shared password-login path. Inline comments below.

Also, a few test gaps versus the ticket's TESTS list: concurrent OTP redeem, concurrent recovery-code redeem, client-bound MFA verification, recovery rate-limit blocking, and rollback/partial-failure after a successful redeem.

app/Http/Controllers/UserController.php
matiasperrone-exo: done

app/Http/Controllers/UserController.php
matiasperrone-exo: The issue is now fixed. `AuthService::loginUser()` (`app/libs/Auth/AuthService.php:445`) now registers the OIDC principal (clear() + register($user->getId(), time())) before Auth::login(), mirroring login(). Note this fix is an uncommitted working-tree change — it was present before this conversation but is not yet in a commit (git status shows M app/libs/Auth/AuthService.php).

  Test coverage - was a real gap, now closed

  The existing tests were not just missing the assertion — two of them were actually erroring (Received getId(), but no expectations were specified), because the fix added a getId() call
  that the tests never accounted for. So the suite was red and nobody had noticed.

  What I changed in tests/unit/AuthServiceValidateCredentialsTest.php:
  1. Promoted $mock_principal_service to a class property so tests can assert on it.
  2. Fixed the two broken loginUser tests (added getId() stubs).
  3. Added testLoginUser_registersPrincipal_withAuthTime — asserts clear() once and register(123, >0) once. This is the regression guard for exactly what caseylocker flagged.
  4. Added testLoginUser_doesNotRegisterPrincipal_whenCannotLogin — asserts the principal is untouched when the user can't log in.
  5. Updated the class docblock to reflect principal registration.

  Results: 7 tests (2 erroring) → 9 tests passing, 51 assertions.

  Mutation-verified: I temporarily reverted the fix and confirmed the new regression test fails (clear expected 1 time, called 0 times), then restored the file byte-for-byte — proving the test genuinely catches the regression rather than passing vacuously.

app/Strategies/MFA/AbstractMFAChallengeStrategy.php
smarcet: @matiasperrone-exo this is an anti pattern , remove it

app/Http/Controllers/UserController.php
```diff
@@ -557,6 +596,287 @@ public function postLogin()
...
+            }
+
+            $user = $this->auth_service->getUserById((int) $pending['user_id']);
+            if (is_null($user) || !$user->isTwoFactorMethodEnabled($method)) {
+                $strategy->clearPendingState();
+                return $this->mfaSessionExpired();
+            }
+
+            // Scope verification to the client the challenge was issued for.
+            $client = $this->resolveClientFromPendingState($pending);
```
smarcet: @matiasperrone-exo 
## Review: client resolution in `verify2FA` duplicates state the memento already owns

`resolveClientFromPendingState` (`app/Http/Controllers/UserController.php:642`) should be replaced with the existing `resolveClientFromMemento()` helper. Findings:

### 1. Duplicated source of truth

The pending-state `client_id` is a session copy of the client that `resolveClientFromMemento()` already resolved at challenge-issue time:

```
postLogin (UserController.php:480)  →  resolveClientFromMemento()
  →  issueMFAChallenge  →  EmailOTPMFAChallengeStrategy::issueChallenge (line 25)
  →  storePendingState(..., $client?->getClientId())   // copies into 2fa_pending_client_id
```

The OAuth2 memento survives the challenge→verify round-trip — `resolveClientFromMemento` re-serializes after `load()` (`UserController.php:628`) and no session regeneration happens until `loginUser` runs *after* verification. The copy adds a second source of truth for a fact the session already holds canonically.

### 2. Internally inconsistent within this PR

- `verify2FA` resolves via pending state (`UserController.php:688`)
- `resend2FA` resolves via `resolveClientFromMemento()` (`UserController.php:853`)

Two endpoints of the same MFA lifecycle disagree on where the client comes from. Worse, resend's path re-runs `issueChallenge` → `storePendingState`, silently **overwriting** the pending-state client with the memento client — so the "snapshot the client at issue time" semantic the comment at `UserController.php:634-637` claims only holds until the first resend.

Concrete drift scenario (two tabs): challenge issued for client A; user starts an authorization request for client B in another tab (memento now = B). Verify succeeds scoped to A, then `login_strategy->postLogin()` (`UserController.php:742`) resumes the **memento** flow — client B. The OTP verification scope and the flow that actually completes disagree. Memento-based resolution makes verify, resend, and post-login continuation agree by construction.

### 3. Silent degradation on stale client

`resolveClientFromMemento` throws `ValidationException("client does not exists")` when the client can't be resolved (`UserController.php:626`); `resolveClientFromPendingState` returns `null` (`UserController.php:648`) — and a null client makes the OTP lookup **unscoped** (`DoctrineOAuth2OTPRepository.php:131` skips the client filter). A deleted/stale client id degrades silently to a wider lookup instead of failing loudly.

### Recommended fix

Swap `UserController.php:688` to `$client = $this->resolveClientFromMemento();`. That makes all of this removable:

- `resolveClientFromPendingState` (`UserController.php:642-649`)
- `KEY_CLIENT_ID` + the `?string $clientId` param on `storePendingState` + `client_id` in `getPendingState`/`clearPendingState` (`AbstractMFAChallengeStrategy.php:16,39,48,72-81`)
- The scoping comments at `EmailOTPMFAChallengeStrategy.php:23-24` and `UserController.php:633-637`

Safety of the swap is already proven in-PR: `resend2FA` calls `resolveClientFromMemento()` mid-MFA and the memento survives for `postLogin` because the helper re-serializes it. `user_id` and `remember` legitimately stay in pending state — they have no memento equivalent. The existing `AbstractMFAChallengeStrategyTest` already calls `storePendingState($userId, $remember)` without a client id, so the signature change is test-compatible.

app/Strategies/MFA/EmailOTPMFAChallengeStrategy.php
smarcet: @matiasperrone-exo remove all the repository->add from everywhere this is an antipattern

smarcet (changes requested):
@matiasperrone-exo review

app/libs/Auth/AuthService.php
```diff
@@ -451,6 +447,13 @@ public function loginUser(User $user, bool $remember): void
         Log::debug("AuthService::loginUser");
         if (!$user->canLogin())
             throw new AuthenticationException("User is not active or cannot login.");
+
+        $this->principal_service->clear();
```
smarcet, line 457: **Minor / non-blocking — pre-existing dead code surfaced while mirroring `login()`.**

Now that `loginUser()` mirrors `login()`'s post-auth principal registration: do **not** also copy `$this->last_login_error = "";` (`login()`, `AuthService.php:396`) into the new methods — that line is dead code:

- It is the **only** reference to `last_login_error` in the entire codebase: no property declaration, no getter, no read anywhere (checked `IAuthService`, controllers, views).
- Because the property is never declared, the assignment creates a **dynamic property** — deprecated since PHP 8.2, so it emits a deprecation notice on every password login under the project's PHP 8.3 runtime.

Since this PR already reworks the login path, deleting that one line in `login()` is a fair drive-by cleanup here; otherwise it can go in a follow-up. Not blocking for this PR.

smarcet: Same category, same file: the import `use Auth\Exceptions\UnverifiedEmailMemberException;` (`AuthService.php:22`) is also unused — it is the only reference to that symbol in the whole file (no code reference, no docblock reference). Safe to drop alongside the `last_login_error` line.

To be clear: only the **import** is dead. The exception class itself is alive — thrown by `CustomAuthProvider.php:132` and caught in `UserController.php:562` — so don't remove the class, just this import.

matiasperrone-exo: Removed "last_login_error" assignment and the import of `Auth\Exceptions\UnverifiedEmailMemberException`.

app/Strategies/MFA/AbstractMFAChallengeStrategy.php
matiasperrone-exo: Removed add() as requested

app/Strategies/MFA/EmailOTPMFAChallengeStrategy.php
matiasperrone-exo: done

app/Http/Controllers/UserController.php
matiasperrone-exo: done

app/Http/Controllers/UserController.php
matiasperrone-exo: `verify2FA` now uses `resolveClientFromMemento()` as `resend2FA`.
`IAuthService::verifyMFAChallenge` now supports a new parameter `?Client`.

matiasperrone-exo: Removed incorrect comments.

app/libs/Auth/AuthService.php
matiasperrone-exo: fixed

caseylocker (approved):
LGTM

